PR #53, opened 2015-01-20
Added left / right functions

nachti:
Extract n left- or right-most element(s)


hadley:
My main concern with this approach is that every other function in `stringr` starts with `str_`, but if you added that prefix `str_left(x, 10)`/`str_right(x, 10)` would be more typing than the equivalent `str_sub(x, end = 10)`, `str_sub(x, -10)`.

nachti:
Ok. I changed it to `str_left` / `str_right` and use stringi now. That's faster. If you use the functions with the default `n` to get just the left- or rightmost character some typing ist saved e.g. `str_left(x)` and it's easier to use in combination with apply & friends.

hadley:
These just don't feel that compelling to me, because they're longer than the existing alternatives. I'd prefer to keep the vocabulary in stringr as small as possible.

PR #57, opened 2015-02-16
Fix existing partial matches in stringr.

craigcitro:
This fixes all the matches I spotted (and adds an extra test, so that we see partial matches in `word` here instead of in dependencies).

There's also some wrapping code that sets and unsets options, so this doesn't immediately arise again -- that should probably move into `testthat`.

### DIFF
--- a/tests/testthat/test-match.r
+++ b/tests/testthat/test-match.r
@@ -1,7 +1,7 @@
 context("Matching groups")
 
 set.seed(1410)
-num <- matrix(sample(9, 10 * 10, rep = T), ncol = 10)
+num <- matrix(sample(9, 10 * 10, replace = T), ncol = 10)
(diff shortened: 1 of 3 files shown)


tests/testthat.R
```diff
@@ -1,4 +1,14 @@
 library(testthat)
 library(stringr)
 
+# TODO(craigcitro): Move this into testthat, and be cleaner about
```
hadley: Would you mind removing this? I'd rather tackle systematically in testthat

craigcitro: the block or the todo?

i'm good either way.

hadley: The whole thing

craigcitro: Done.

hadley:
Thanks!

PR #62, opened 2015-03-30
Use full arg name for rep(length.out=...)

jcheng5:
This bug causes warnings if options(warnPartialMatchArgs=TRUE).


jcheng5:
You can see this warning when you do devtools::install_github.

hadley:
Thanks!

PR #76, opened 2015-05-29
Another try at str_interp

smbache:
Not an expert at updating pull requests, so ended up making a new one.Your comments should be fixed.


hadley:
Easiest way to update is to make a branch for the PR - then when you update the branch, the PR updates automatically.

hadley:
Thanks!

PR #77, opened 2015-05-31
Bugfix + small feature

smbache:
This update does two things: 

1. `str_interp` would fail when no placeholders were found in the input string. This is fixed.
2. Added a suggested feature that makes it easier to break a long single-line string in multiple lines of source code without needing to use e.g  `paste`. Instead if the input string is a formula, hyphens (dashes) can be used to concatenate long strings: 

```R
max_char <- 80
str_interp(~"This particular line is so long that it is hard to write "-
            "without breaking the ${max_char}-char barrier!")
```

Here `str_interp` uses an internal utility function `unwrap_string_formula` which will recognize `-` as a hyphen (and not as a minus).

### DIFF
--- a/R/interp.R
+++ b/R/interp.R
@@ -34,16 +34,52 @@
 #' # Or a data frame
 #' str_interp("Values are $[.2f]{max(Sepal.Width)} and $[.2f]{min(Sepal.Width)}.",
 #'            iris)
+#'
+#' # Use a formula and hyphens when the string is long
+#' max_char <- 80
+#' str_interp(~"This particular line is so long that it is hard to write "-
+#'   "without breaking the ${max_char}-char barrier!")
 str_interp <- function(string, env = parent.frame())
 {
+
+  if ("formula" %in% class(string))
+    string <- unwrap_string_formula(string)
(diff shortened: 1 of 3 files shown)


hadley:
Would you mind including tests for those features please?

hadley:
(I really need to turn on code coverage)

smbache:
Sure. 

In the meantime; which is better, a plus or a hyphen (or either) for the above feature?

smbache:
Added the tests.

smbache:
What would you want to happen in the latter, where the expression is not a valid R expression? expect_error? And what is the idea for the second? That it is not a complete placeholder?

gagolews:
I think that a clean error (parse error?) should be raised. What do you think?
And in the first case, there should be no crash.

And what about `${aaa   bbbb}`? Also a parse error, `aaa bbbb` is not a valid name.
And also: `${aaa${bbb}ccc}` may be problematic

smbache:
I think they give rather readable errors. Nested use of placeholders did not produce a nice error, but now this is checked and provides a better error. In general nested use is *never* allowed.

Issue #68, opened 2015-05-12
R CMD check failed: The requested ICU resource file cannot be found


yihui:
Travis has been failing but your Travis script failed to capture the failure for some reason: https://travis-ci.org/hadley/stringr/builds/61152178 I noticed it because my knitr repo started to fail after stringr was upgraded (https://travis-ci.org/yihui/knitr/jobs/61492954): 

> The requested ICU resource file cannot be found. Possible problem: ICU data has not been downloaded yet. Call `stri_install_check()`. (U_FILE_ACCESS_ERROR)

stringr has the same error, which I don't completely understand. This may be related to #52.

BTW, you most recent check also failed (for a different reason): https://travis-ci.org/hadley/stringr/builds/61467874 and Travis failed to capture it, either.

gagolews:
This is almost identical to https://github.com/Rexamine/stringi/issues/155

hadley:
@yihui I think the failure might be intermittent

yihui:
@gagolews Oh I did not realize it had been reported. Thanks. But I do not think we have got to the bottom of this issue. It is not necessary to install stringi from Github, either. You just have to install it from source (e.g. the CRAN source package works), and the binary package from marutter's PPA does not work. Well, my experience is that there are quite a few broken binary packages in marutter's PPA...

maxheld83:
I get the sample problem, irrespective of whether `stringr` is loaded from github or CRAN:

```
>     plots <- array.viz(QmethodRes = results)  # produces plots list
Error in stri_trans_nfc(str) : 
  The requested ICU resource file cannot be found. Possible problem: ICU data has not been downloaded yet. Call `stri_install_check()`. (U_FILE_ACCESS_ERROR)
Calls: array.viz -> str_wrap -> stri_wrap -> stri_trans_nfc -> .Call
Execution halted
```

+1 for 
> fix the root cause

Ps.: is it just me or is the 3k-line log from Travis.CI even more confusing than ordinary R CMD check?

jeroenooms:
This is still not resolved. The problem is that `stringi` fails to detect `libicu-dev` on Ubuntu and instead tried to build a bundled version, which doesn't work very well.

jeroenooms:
I have proposed a solution here: https://github.com/Rexamine/stringi/issues/162

gagolews:
Maybe let's first try with debugging the Ubuntu libicu-dev-related issue?

gagolews:
[...]

PR #82, opened 2015-06-13
Update travis conf

gagolews:
Hadley, I think I now have a fix for #68. Anyway, as stringr and stringi are now so tightly connected with each other, what about travis-testing stringr against the github version of stringi? I updated `.travis.yml` accordingly for that.


hadley:
Thanks!

Issue #83, opened 2015-06-25
str_replace errors when using "$" as a replacement.


dkesh:
The code below works for me on some boxes, but not others and I am attempting to figure out what the difference is:

```
> stringr::str_replace("aba", "b", "$" )
Error in stri_replace_all_regex(x, c("\\$", "\\\\(\\d)"), c("\\\\$", "\\$$1"),  : 
  Unknown ICU error. (U_REGEX_INVALID_CAPTURE_GROUP_NAME)

Enter a frame number, or 0 to exit   

1: stringr::str_replace("aba", "b", "$")
2: fix_replacement(replacement)
3: stri_replace_all_regex(x, c("\\$", "\\\\(\\d)"), c("\\\\$", "\\$$1"), vecto
```

dkesh:
stringr version 1.0.0 and stringi version 0.5.2 on both working and non-working boxes.

dkesh:
On the boxes that fail (both Ubuntu and OS X):

```
> stringi::stri_install_check()
stringi_0.5.2; en_US.UTF-8; ICU4C 55.1; Unicode 7.0
All tests completed successfully.
```

On the box that works:
```
> stringi::stri_install_check()
stringi_0.5.2; en_US.UTF-8; ICU4C 52.1; Unicode 6.3
All tests completed successfully.
```

gagolews:
`$` in a replacement string (in a regex search task) has a special meaning - it refers to a capture group. It should be followed by an integer. So to use `$`, it shall be escaped.

```
stringr::str_replace("aba", "b", "\\$" )
```

you may also want to  consider using `fixed("b")` as a search patter in this example.

From syntactic view point, this was an error too in ICU 52.1, but apparently in ICU developers decided to throw an error and no longer ignore it in a later ICU release (like 55).

In other words, this is not a bug.

dkesh:
Running the escaped example returns this:

   > stringr::str_replace( "aba", "b", "\\$")
    [1] "a\\$a"

Perhaps stringr:::fix_replacement is no longer necessary on newer versions of ICU?

dkesh:
The extra backslash is missing because I forgot to escape it for markdown.  That should read:

    > stringr::str_replace( "aba", "b", "\\\\$")
    [1] "a\\\\$a"

dkesh:
The second backslashes in each of the code line got gobbled up by markdown; just imagine they are there.

gagolews:
[...]

PR #84, opened 2015-06-27
BUGFIX for fix_replacement

gagolews:
This is a bugfix for #83.


gagolews:
This works both for ICU 52 and ICU 55.

hadley:
Thanks!

PR #94, opened 2015-09-23
Add str_match to "See Also" of str_extract

### LINKED ISSUE #93: Documentation: str_extract help page should mention str_match in "See Also"
I was looking for the function to extract capture groups from a regex match, and my first guess was str_extract. I quickly figured out from the help page that this was the wrong choice though. Eventually, I found that the correct one was str_match. But it would have been a lot easier to find if str_extract had led me there by mentioning it in the "See also" section. I think a lot of people probable make the same mistake, finding str_extract when they're looking for str_match, since it's not clear from the names alone which one returns capture groups and which one only returns the complete match.

DarwinAwardWinner:
This completes the symmetry of str_match referring to str_extract in its
own "See Also" section.

Fixes #93.

### DIFF
--- a/man/str_extract.Rd
+++ b/man/str_extract.Rd
@@ -54,6 +54,7 @@ str_extract_all(shopping_list, "\\\\b[a-z]+\\\\b", simplify = TRUE)
 str_extract_all(shopping_list, "\\\\d", simplify = TRUE)
 }
 \seealso{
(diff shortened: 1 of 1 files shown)


hadley:
Like the comment at the top of the file says, you need to edit the source roxygen comment :/

PR #136, opened 2016-10-31
@seealso link to str_subset() from str_detect()

t-kalinowski:
Hello,

Minor doc improvement. I always forget what `str_subset()` is called, and end up looking at the docs for `str_detect()` first when I want it, then when it's not there I end up clicking randomly through all the `str_*` functions till I find the one I'm looking for.


hadley:
Thanks!

PR #151, opened 2017-03-09
Create pipe_p.R

duarteguilherme:
I'm Not sure if someone uses a pipe to emulate + python function to joining strings.

I guess infix operators are more useful for this type of operation.


duarteguilherme:
Sorry for that. I've just seen someone proposed the same solution "%f%" below.

Anyway, I think this could me integrated to the function. 

%s+% from stringi is kind of verbose.

hadley:
Thanks for the pull request, but currently I don't want to include infix string pasting in string.

PR #158, opened 2017-04-24
Feature request: vectorise str_interp in interpolated expressions

mgirlich:
I like the handling of vectors in `paste()` a lot in that it cycles over all the elements of the input expressions:

```
paste(month.abb[1:2], "is the", 1:2, "month of the year.")
# "Jan is the 1 month of the year." "Feb is the 2 month of the year."
```

I think the behaviour of `str_interp()` is less useful and not intuitive here:

```
str_interp("${month.abb[1:2]} is the ${1:2} month of the year.")
# "Jan is the Feb month of the year."
```

One should at least get a warning or rather an error for this unexpected behaviour.
With the pull request one gets instead

```
str_interp("${month.abb[1:2]} is the ${1:2} month of the year.")
# "Jan is the 1 month of the year." "Feb is the 2 month of the year."
```

As this breaks the current behaviour one might want to add an extra parameter for allowing this.

### DIFF
--- a/R/interp.R
+++ b/R/interp.R
@@ -69,14 +69,15 @@ str_interp <- function(string, env = parent.frame()) {
     # check that lengths are valid
     lengths <- sapply(replacements, length)
     max_length <- max(lengths)
-    if (!all(lengths %in% c(1, max_length))) {
+    if (!all(lengths == 1L | lengths == max_length)) {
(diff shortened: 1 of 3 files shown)


hadley:
I'd say `str_interp` is soft-deprecated in favour of the glue package. We want be making any changes to `str_interp()` in the future.

PR #165, opened 2017-05-12
Allow str_trunc to pass through NA's in input.

### LINKED ISSUE #162: str_trunc can't handle NA's in input
Passing a single `NA` into `str_trunc` leaves it untouched, as expected

```r
NA_character_ %>% str_trunc(10)
```

But a vector with one or more NA's throws an error

```r
c("foo bar baz", NA) %>% str_trunc(10)
c(NA_character_, NA_character_) %>% str_trunc(10)
```
> Error in string[too_long] <- switch(side, right = str_c(str_sub(string[too_long],  : 
  NAs are not allowed in subscripted assignments

which happens without piping, as well.

It seems like the expected behaviour would be to truncate the "real" strings, and leave the NA's untouched, rather than throw an error.

## Session Info
```
> devtools::session_info()
Session info ----------------------------------------------------------------------------------------------------------------------------
 setting  value                       
 version  R version 3.4.0 (2017-04-21)
 system   x86_64, linux-gnu           
 ui       RStudio (1.0.143)           
 language (EN)                        
 collate  en_US.UTF-8                 
 tz       America/New_York            
 date     2017-05-09                  

Packages --------------------------------------------------------------------------------------------------------------------------------
 package  * version date       source        
 devtools   1.12.0  2016-12-05 CRAN (R 3.4.0)
 digest     0.6.12  2017-01-27 CRAN (R 3.4.0)
 magrittr * 1.5     2014-11-22 CRAN (R 3.4.0)
 memoise    1.1.0   2017-04-21 CRAN (R 3.4.0)
 stringi    1.1.5   2017-04-07 CRAN (R 3.4.0)
 stringr  * 1.2.0   2017-02-18 CRAN (R 3.4.0)
 withr      1.0.2   2016-06-20 CRAN (R 3.4.0)
```
(Not the dev version, but git blame suggests this function hasn't changed in two years)

ClaytonJY:
One-line change to function code, with two new tests to prevent regression.

Fixes #162.

### DIFF
--- a/R/pad-trim.r
+++ b/R/pad-trim.r
@@ -55,6 +55,8 @@ str_trunc <- function(string, width, side = c("right", "left", "center"),
   side <- match.arg(side)
 
   too_long <- str_length(string) > width
+  too_long[is.na(too_long)] <- FALSE
+
   width... <- width - str_length(ellipsis)
 
   string[too_long] <- switch(side,
--- a/tests/testthat/test-trunc.r
+++ b/tests/testthat/test-trunc.r
@@ -1,0 +1,12 @@
+context("Test truncation")
+
+test_that("NA values in input pass through unchanged", {
+  expect_equal(
+    str_trunc(NA_character_, width = 5),
+    NA_character_
+  )
+  expect_equal(
+    str_trunc(c("foobar", NA), 5),
+    c("fo...", NA)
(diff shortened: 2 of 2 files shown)


hadley:
Thanks!

PR #174, opened 2017-07-07
duplicated examples in str_replace

cderv:
Example was a bit long but it is because there were some identical lines. 
I deleted the duplicated example lines


hadley:
Thanks!

PR #178, opened 2017-08-02
Added str_remove() and str_remove_all()

Shians:
This is just an alias for str_replace(string, pattern, "") but I feel like it's a common enough usage to deserve its own functions.


hadley:
Can you please add a bullet to NEWS? It should briefly describe the change (starting with name of the function), and crediting yourself with `(@yourname, #issuenumber)`. 

Thanks!

Shians:
Thanks Hadley, I've updated the NEWS.

hadley:
Thanks!

PR #180, opened 2017-08-21
Update stringr.Rmd

nokome:
Fix a single small typo


hadley:
Thanks - someone else spotted it too.

PR #181, opened 2017-09-06
Add link to tidyverse/glue

jameshowison:
I spent quite a bit of time trying to make str_interp work like tidyverse/glue_data ... my logic was "tidyverse string stuff is in stringr, str_interp sounds right." So just as tidyverse/glue points to str_interp as an option, this points back (I hope, I'm afraid my rdoc is very weak).


hadley:
Thanks! We'll probably eventually deprecated `str_interp()` in favour of `glue`

PR #184, opened 2017-10-23
stringr vignette: replace sub_str with str_sub

florisvdh:


hadley:
Thanks - you're the third person to spot this mistake!

PR #185, opened 2017-10-25
regexp vignette: correct possessive quantifiers

florisvdh:


hadley:
Doh! Thanks for fixing this

PR #187, opened 2017-10-27
Eliminate a partial arg match

jennybc:
Improves life for those of us living the `warnPartialMatch` lifestyle (https://github.com/r-lib/usethis/issues/64)


hadley:
Thanks!

PR #190, opened 2017-11-06
Add "use_length" argument to str_pad()

### LINKED ISSUE #188: str_pad count 1 chinese char as 2
> vchs= "中,中中,中中中"
> yc=str_split(vchs,",",simplify = T)
> print(yc)
     [,1] [,2]   [,3]    
[1,] "中" "中中" "中中中"
> yl=str_length(yc)
> print(yl)
[1] 1 2 3
> print(max(yl))
[1] 3
> ycd=stringr::str_pad(yc,width=max(str_length(yc)),side = "right",pad=" ")
> ycd
[1] "中 "    "中中"   "中中中"
> ycd=stringr::str_pad(yc,width=2*max(str_length(yc)),side = "right",pad=" ")
> ycd
[1] "中    " "中中  " "中中中"

yutannihilation:
Fixes #188 

As I commented on #188, `stri_pad_*()` gains `use_length` (https://github.com/gagolews/stringi/issues/149) to control whether to use the total code point width or the number of code points as "width" of a string. But, the corresponding argument is not added on stringr's side so far. This PR simply adds it to `str_pad()`.

As base R's `nchar()` has similar argument `type` which takes either `"bytes"`, `"chars"` (default), or `"width"`, I guess there must be some people who wants to control this behaviour, so exposing this argument is reasonable.


hadley:
I think this is a good idea, but the docs are confusing me. Does this change the current default behaviour? (It probably should since you're usually padding to make strings aligned for output)

tests/testthat/test-pad.r
```diff
@@ -18,3 +18,8 @@ test_that("directions work for simple case", {
   expect_equal(pad("left"),   "       had")
   expect_equal(pad("both"),   "   had    ")
 })
+
+test_that("padding based of length works", {
+  expect_identical(str_pad("\u4e2d", width = 6, side = "both"),                    "  \u4e2d  ")
```
hadley: Needs comment describing what "\u4e2d" is

yutannihilation:
> Does this change the current default behaviour?

No, but I think you don't need to be afraid.

The default behaviour already had been changed on the stringi's side after version [0.5-2](https://github.com/gagolews/stringi/releases/tag/v0.5-2), which pads strings based on the total character widths so that strings are properly aligned for output. Users who want the old behaviour can still choose the old behaviour by `use_length = TRUE` ("length" means "the number of characters" here, which may be a bit confusing...).

Does my explanation make sense? I'm sorry maybe my English is not good enough to document this nicely :(

hadley:
In that case, this doesn't seem super useful to me given that the point of `str_pad()` is to align things visually.

yutannihilation:
Agreed. This is not super useful. Yet, it seems some people like #188 need the behaviour consistent with the default of base `nchar()` (I don't know the exact use cases, though).

Maybe you can close this PR for now and reopen if needed  :)

yutannihilation:
Thanks for merging!

PR #205, opened 2018-01-04
Clean roxygen doc in str_glue and vignette

dchiu911:
- There is no `glue::data`. The correct inheritance should be from `glue::glue_data`, which is already being passed on L39.
- Spelling typo in release 1.2.0 article. `pkgdown` needs to be run by a maintainer for the tidyverse website to be updated.
- Reflowed comment for `str_c` description to trigger a travis build, which now passes all checks


hadley:
Thanks!

PR #206, opened 2018-01-12
Fixed unexpected behavior of str_trunc

### LINKED ISSUE #203: Unexpected behavior of str_trunc
I encountered an unexpected behavior of `str_trunc()`, here is a reprex of my problem.

In stringr v1.2.0 the result is as expected:
```r
str_trunc(c('abcd', 'abcde', 'abcdef'), width = 5)
[1] "abcd"  "abcde" "ab..."
```

In stringr v1.2.0.9000 the result not what one would unexpect:
```r
str_trunc(c('abcd', 'abcde', 'abcdef'), width = 5)
[1] "abcd"   "abcde"  "abcdef"
```

And changing the order of the elements of the vector also impacts the output, which leads me to believe that the truncation is performed or not based on the number of characters in the first element of the vector of strings
```r
str_trunc(c('abcdef', 'abcde', 'abcd'), width = 5)
[1] "ab..." "ab..." "ab..."
```

austin3dickey:
Fixes #203.

With the NA logic fix to str_trunc 2 months ago, "&&" was used instead of "&", causing str_trunc to only work on the first element of the vector.

I added a test as well, but nothing in the NEWS since this change wasn't ever released.

### DIFF
--- a/R/trunc.R
+++ b/R/trunc.R
@@ -18,7 +18,7 @@ str_trunc <- function(string, width, side = c("right", "left", "center"),
                       ellipsis = "...") {
   side <- match.arg(side)
 
-  too_long <- !is.na(string) && str_length(string) > width
+  too_long <- !is.na(string) & str_length(string) > width
   width... <- width - str_length(ellipsis)
 
   if (width... < 0) stop("`width` is shorter than `ellipsis`", .call = FALSE)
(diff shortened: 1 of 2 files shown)


hadley:
Can you please add a bullet to NEWS? It should briefly describe the change (starting with name of the function), and crediting yourself with `(@yourname, #issuenumber)`.

hadley:
Actually, I'll add the news.

PR #212, opened 2018-01-30
Include see also pointer to tidyr::extract

jameshowison:
I turned to str_match and str_extract for doing what tidyr::extract does (and tidy::extract is a much more convenient form!). So I added a cross-link to the see also.

Thanks to markdly at community.rstudio.org showing me the right way! https://community.rstudio.com/t/getting-parts-of-string-using-regex-and-capturing-parens-in-mutate

### DIFF
--- a/R/match.r
+++ b/R/match.r
@@ -11,7 +11,8 @@
 #'
 #' @seealso [str_extract()] to extract the complete match,
 #'   [stringi::stri_match()] for the underlying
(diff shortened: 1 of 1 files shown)


hadley:
Unfortunately we can't do this because R CMD check will complain that we're linking to a package that's not listed in the DESCRIPTION 😢

PR #214, opened 2018-02-02
Str_dice

ewallace:
New function str_dice breaks down a string into fixed-length substrings. This is very useful in bioinformatics for breaking a coding sequence of DNA into codon-sized triplets.

I did not find this functionality in stringr or stringi already, but str_dice is a simple wrap of stri_sub using seq to select the positions automatically.

Hope it's useful.
Edward


hadley:
The combination of the specificity and the simplicity of the implementation means that this doesn't feel right in stringr to me.

PR #224, opened 2018-03-05
new function to standardize strings

pachevalier:
This pull request introduce a new function `str_standardize()` which makes it easy to remove accents and non ascii characters from a string. 

This is very useful for dealing with non English data such as French data.

One very useful use is to standardize names in a data frame (see [`set_standard_names()`](https://github.com/pachevalier/tricky/blob/master/R/tibbles.R#L68) in my personal package.


pachevalier:
I've opened an issue in the readr package  : https://github.com/tidyverse/readr/issues/826

PR #247, opened 2018-07-23
Add markdown output to str_view

jrnold:
Add an option to output markdown format from `str_view` rather than just the HTML widget format. This is useful for outputting to LaTeX, or if a simpler format is needed.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,4 +30,4 @@ Suggests:
 VignetteBuilder: knitr
 LazyData: true
 Roxygen: list(markdown = TRUE)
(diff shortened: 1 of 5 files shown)


hadley:
I'm not sure I want to make `str_view()` any more complicated. It might be better to submit this to [regexplain](https://github.com/gadenbuie/regexplain) ?

jrnold:
No problem; that makes total sense. The issue for me was that the output of `str_view` did not play nicely with PDF in the R4DS solutions. 

Thanks for pointing me to regexplain!

PR #250, opened 2018-08-07
str_explode: split into single characters

jonocarroll:
Simple helper function to split a string into single characters. Not sure why this isn't a helper in base, I've seen the use-case come up more than a few times.

Some thought could go into whether or not to support both `list` and `vector` inputs for vectorisation - `base::strsplit(string, "")` always returns a `list` which is annoying if only the first element is ever needed (most cases I've seen) resulting in a lot of `strsplit(string, "")[[1]]`. This implementation at least preserves the input type.

``` r
library(stringr) # this PR

str_explode("apple")
#> [1] "a" "p" "p" "l" "e"

str_explode(c("apple", "banana"))
#>  [1] "a" "p" "p" "l" "e" "b" "a" "n" "a" "n" "a"

str_explode(list("apple", "banana"))
#> [[1]]
#> [1] "a" "p" "p" "l" "e"
#> 
#> [[2]]
#> [1] "b" "a" "n" "a" "n" "a"

### LESS TRIVIAL EXAMPLE:
set.seed(1)
(v <- stringi::stri_rand_lipsum(1))
#> [1] "Lorem ipsum dolor sit amet, habitant at ullamcorper eu nunc nisl, sed nisl semper. Turpis dolor, interdum augue, sit. Venenatis lacinia posuere ac fermentum leo euismod tristique pretium, lectus urna congue sed. Risus scelerisque duis risus praesent mi vitae ut et. Tincidunt, metus nullam eros himenaeos, vestibulum. Lorem donec tristique quis risus ornare ut fusce donec. Placerat turpis donec risus per bibendum velit condimentum nascetur mi. Iaculis erat ipsum facilisis vel aenean ac vitae non vitae. In maecenas fusce mollis suscipit lacus leo."

## table of character appearances
v %>% 
  tolower() %>% 
  str_explode() %>% 
  table()
#> .
#>     ,  .  a  b  c  d  e  f  g  h  i  l  m  n  o  p  q  r  s  t  u  v 
#> 82  7  9 30  4 22 13 57  4  2  2 47 24 23 31 21 12  4 29 45 34 41  7
```

Created on 2018-08-07 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).


jonocarroll:
Tests can certainly be added to satisfy `codecov` if this is supported.

hadley:
IMO it's better for to teach about why `[[1]]` is the right thing to do here.

PR #259, opened 2018-10-01
Add negate argument to str_subset(), str_detect(), and str_which()

yutannihilation:
Currently, there's no cool way of doing `grep -v` with stringr. If I want to negate the pattern, I need to use `str_detect()`, but this is not handy.

``` r
library(stringr)

fruit <- c("apple", "banana", "pear", "pinapple")
fruit[!str_detect(fruit, "^p)]
```

On the other hand, stringi's equivalent functions have `negate` argument since version 1.1.1. So, it seems easy to introduce the argument in stringr package. Note that my primary motivation was `str_subset()`, only, but as `stringi::stri_detect()` also has `negate`, I included the equivalent functions, `str_detect()` and `str_which()`.

``` r
library(stringr)

fruit <- c("apple", "banana", "pear", "pinapple")
str_subset(fruit, "^p")
#> [1] "pear"     "pinapple"

str_subset(fruit, "^p", negate = TRUE)
#> [1] "apple"  "banana"
```

<sup>Created on 2018-10-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


hadley:
Thanks!

PR #261, opened 2018-10-08
Is this example correct?

osorensen:
I do not get where yz and xyz are coming from in this example. Proposing a fix which I believe is correct.


hadley:
Yes, it's correct - it's sample input to the regular expression.

PR #273, opened 2019-01-19
Add requireNamespace so str_view() fails more gracefully 

### LINKED ISSUE #263: Fail more gracefully when str_view() is called in the absence of htmlwidgets
Although htmlwidgets is in Suggests, we don't quite follow our own best practice within `str_view()` and related helpers re: checking if it's installed. The error message is sub-optimal:

``` r
> str_view(fruit, "melon")
Error in loadNamespace(name) : there is no package called 'htmlwidgets'
```

It should be more precise about the problem and the solution.

Resources to consult:

  * Consult the [Namespaces chapter](http://r-pkgs.had.co.nz/namespace.html) of R Packages, specifically the bit about using `requireNamespace("x", quietly = TRUE)`
  * Search for "requireNamespace" on GitHub where the user is "tidyverse" or "r-lib" for examples of how it's done in other packages. Enter something like this into GitHub search: `requireNamespace user:r-lib`. Focus on the hits found in "Code".
  * Read the section of the tidyverse style guide on [error messages](http://style.tidyverse.org/error-messages.html).

*This issue was written in preparation for [Tidyverse Developer Day](https://www.tidyverse.org/articles/2018/08/tidyverse-developer-day/). Unless there's a really good reason to do otherwise, please leave it for handling then.*

AmeliaMN:
in the absence of htmlwidgets

To fix #263

### DIFF
--- a/R/view.R
+++ b/R/view.R
@@ -74,9 +74,9 @@ str_view_widget <- function(lines) {
   )
   html <- htmltools::HTML(bullets)
 
-if(!requireNamespace("htmlwidgets", quietly = TRUE)){
-  message("htmlwidgets package required for str_view(). \nPlease install.packages(\"htmlwidgets\") to use this functionality.")
-}else{
+  if (!requireNamespace("htmlwidgets", quietly = TRUE)) {
+    stop("htmlwidgets package required for str_view(). \nPlease install.packages(\"htmlwidgets\") to use this functionality.")
+  }
   size <- htmlwidgets::sizingPolicy(
     knitr.figure = FALSE,
     defaultHeight = pmin(10 * length(lines), 300),
@@ -89,5 +89,4 @@ if(!requireNamespace("htmlwidgets", quietly = TRUE)){
     sizingPolicy = size,
     package = "stringr"
   )
-}
 }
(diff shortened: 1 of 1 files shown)


R/view.R
```diff
@@ -74,6 +74,9 @@ str_view_widget <- function(lines) {
   )
   html <- htmltools::HTML(bullets)
 
+if(!requireNamespace("htmlwidgets", quietly = TRUE)){
```
hadley: Can you please use styler to restyle this block?  (You're missing a few spaces)

R/view.R
```diff
@@ -74,6 +74,9 @@ str_view_widget <- function(lines) {
...
+  message("htmlwidgets package required for str_view(). \nPlease install.packages(\"htmlwidgets\") to use this functionality.")
```
hadley: Oh and this should be a stop, and hence you don't need the `else` block

R/view.R
```diff
@@ -74,6 +74,9 @@ str_view_widget <- function(lines) {
   )
   html <- htmltools::HTML(bullets)
 
+  if (!requireNamespace("htmlwidgets", quietly = TRUE)) {
+    stop("htmlwidgets package required for str_view(). \nPlease install.packages(\"htmlwidgets\") to use this functionality.")
```
hadley: Can you please add `call. = FALSE`?

hadley:
Thank you!!

Issue #276, opened 2019-02-02
Documentation on str_c()


paleolimbot:
In teaching a "basic R" tutorial the other day, I had people use `str_c()` as a quick example of a vectorized operation on a character vector, and it became very clear in a room of beginner R users that the help file does little to explain what the function does (especially for beginners).

<img width="744" alt="screen shot 2019-02-02 at 11 51 28 am" src="https://user-images.githubusercontent.com/10995762/52166230-fe26db80-26e0-11e9-8898-8601be2d57d8.png">

Perhaps adding a line like `paste()` has at the front like "Concatenate vectors after converting to character" might be helpful? I am happy to PR on this as well (should I have just PRed or is it better to open an issue first?).

hadley:
Yeah, a couple of sentences would be great.

(For obvious doc fixes like this, it would be fine to go straight to a PR)

PR #277, opened 2019-02-02
Add an alternative description for `str_c`

paleolimbot:
See issue #276, the initial explanation of what `str_c()` does is a little intimidating (or it was for a few folks in a class I taught recently). I would welcome any edits to what I've come up with...it is a hard thing to explain! If `str_collapse()` were a different function it may be easier to explain, but I get that `str_c()` is trying to be a drop-in replacement for `paste0()`.

### DIFF
--- a/R/c.r
+++ b/R/c.r
@@ -1,5 +1,10 @@
 #' Join multiple strings into a single string.
 #'
+#' Joins two or more vectors element-wise into a single character vector,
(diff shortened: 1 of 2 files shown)


hadley:
I like it! 

If I was to implement `str_c()` today, I'd definitely not include the `collapse` argument, saving that for a different function.

Issue #245, opened 2018-06-25
Unconsistent behavior in `word` function


pgrandinetti:
I don't get if the following behavior of the `word` function is desired, or is a bug

```
> words = "this is a very long sentence to test"
> word(words, start=-4, end=-1)  # last 4 words, OK
[1] "long sentence to test"
> word(words, start=-8, end=-1)  # OK
[1] "this is a very long sentence to test"
> word(words, start=-9, end=-1)  # OK, imho
[1] NA
> word(words, start=-10, end=-1)  # what is this?!
[1] "is a very long sentence to test" "a very long sentence to test"    "very long sentence to test"     
[4] "long sentence to test"           "sentence to test"                "to test"                        
[7] "test"    
```

hadley:
Could you please rework your reproducible example to use the [reprex package](http://reprex.tidyverse.org) ? That makes it easier to see both the input and the output, formatted in such a way that I can easily re-run in a local session.

pgrandinetti:
I guess we just need to clarify what is the behavior of the `word` function when the input argument `start` is negative and is greater than the sentence length.

From your doc:

> **start**:  integer vector giving position of first word to extract. Defaults to first word. If negative, counts backwards from last character.

But the behavior is not always clear. Sometimes I get a list as result.

If you copy this and run `reprex()` should give what you asked for (assuming that I understood what you meant!)
```
library(stringr)
library(reprex)
sentence = "this is a very long sentence to test"
word(sentence, start=-8, end=-1)  # OK
word(sentence, start=-9, end=-1)  # OK, imho because 9 > length
word(sentence, start=-10, end=-1)  # what is this, a vector?!
```

hadley:
Please include the _output_ of using reprex in the issue so I can see both the input and the output.

pgrandinetti:
```
library(stringr)
library(reprex)
sentence = "this is a very long sentence to test"
word(sentence, start=-8, end=-1)  # OK
#> [1] "this is a very long sentence to test"
word(sentence, start=-9, end=-1)  # OK, imho because 9 > length
#> [1] NA
word(sentence, start=-10, end=-1)  # what is this, a vector?!
#> [1] "is a very long sentence to test" "a very long sentence to test"   
#> [3] "very long sentence to test"      "long sentence to test"          
#> [5] "sentence to test"                "to test"                        
#> [7] "test"
```

pdelboca:
[...]

PR #282, opened 2019-02-10
Fix bug in word()

pdelboca:
Fixing unconsistent behavior described in #245 and adding some test.


tests/testthat/test-word.r
```diff
@@ -10,3 +10,9 @@ test_that("word extraction", {
 test_that("words past end return NA", {
   expect_equal(word("a b c", 4), NA_character_)
 })
+
+test_that("negative parameters", {
+  expect_equal("moon",word("walk the moon", -1, -1))
```
hadley: Can you please add the missing space before `word()`?

Also you only need the first line here as the other two test the same code path.

pdelboca: Done! 

Although I will suggest to keep the other two lines since they test a specific behavior: return all the sentence if the negative start parameter is greater than the number of words. Important to have a consistent API with `str_sub`.

Let me know if you want to remove it anyways!

pdelboca:
Made all the changes. Let me know if anything else is needed!

hadley:
Thanks!

PR #283, opened 2019-02-18
Add new `replace` parameter in `str_trunc`

denrou:
This parameter allows the ellipsis to be added after truncation is done, i.e. without being taken into account in the `width` parameter.

## Example

``` r
library(stringr)
library(dplyr)
#> 
#> Attachement du package : 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(glue)
#> 
#> Attachement du package : 'glue'
#> The following object is masked from 'package:dplyr':
#> 
#>     collapse

set.seed(123)
df <- tibble(value = sample(0:9, 30, replace = TRUE), group = sample(c("a", "b", "c"), 30, replace = TRUE))

df %>% 
  group_by(group) %>% 
  summarise(value = str_trunc(toString(value), 7, ellipsis = glue(" ... + {length(value) - 1} more."), replace = FALSE))
#> # A tibble: 3 x 2
#>   group value                 
#>   <chr> <chr>                 
#> 1 a     9, 8, 5 ... + 12 more.
#> 2 b     0, 4, 6 ... + 7 more. 
#> 3 c     2, 7, 4 ... + 8 more.
```

<sup>Created on 2019-02-18 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: stringr
 Title: Simple, Consistent Wrappers for Common String Operations
-Version: 1.4.0.9001
+Version: 1.4.0.9002
 Authors@R: 
     c(person(given = "Hadley",
              family = "Wickham",
@@ -36,4 +36,4 @@ Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 6.1.1
-Date: 2019-02-18
+Date: 2019-02-19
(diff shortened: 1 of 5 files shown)


hadley:
This seems a bit specialised to me, sorry (and you can easily manually adjust the width currently)

Issue #266, opened 2018-12-10
Vignette comparing stringr to base R


hadley:
Starting with a table lets you easily find the stringr function equivalent to a base R function (and vice versa)

moodymudskipper:
Great idea, it's also not always clear why an equivalent exists or how it differs from a *`stringi`* close equivalent, I often found myself wondering:

* Is it just a matter of reordering the parameters ?
* Is it faster ?
* Does it address differently corner cases ?

Usually the doc of stringr and stringi contain the answer but it takes time to dig it out.

I did some work on SO here in case it helps :

[Difference between `paste`, `str_c`, `str_join`, `stri_join`, `stri_c`, `stri_paste`?](https://stackoverflow.com/questions/53118271/difference-between-paste-str-c-str-join-stri-join-stri-c-stri-pa/53118273#53118273)

And I also had posted comparisons with base functions in the examples off rdocumentation.com : 

[str_detect](https://www.rdocumentation.org/packages/stringr/versions/1.3.1/topics/str_detect)
[str_length](https://www.rdocumentation.org/packages/stringr/versions/1.3.1/topics/str_length)

hadley:
I don't think a comparison with stringi would be terribly useful as the functions are very close.

PR #284, opened 2019-02-19
Compare to base vignette

sastoudt:
This pull request includes a vignette comparing base to stringr. (https://github.com/tidyverse/stringr/issues/266)

It starts with tables mapping stringr functions from the [stringr cheat sheet](https://github.com/rstudio/cheatsheets/blob/master/strings.pdf) to the analogous base R functions. It then continues by reproducing the examples from the [stringr documentation](https://cran.r-project.org/web/packages/stringr/stringr.pdf) in base R.

Note: I was not able to get the output of `stringr::str_wrap()` and `strwrap()` to match.


gagolews:
I'd be happy if stringi was mentioned too

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
+---
+title: "Comparing stringr to base R"
+author: "Sara Stoudt"
+date: "`r Sys.Date()`"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Vignette Title}
```
hadley: This needs an update too

hadley:
@gagolews maybe that's a task for a separate vignette? I think a stringr vs stringi vignette would focus on how stringi is much finer grained that stringr, and has more features. I _think_ it'd be hard to do that and compare stringr to base R in a single vignette.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
...
+```{r setup, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+This vignette aims to compare stringr functions to base R equivalents. Following the groupings of functions from the [stringr cheat sheet](https://github.com/rstudio/cheatsheets/blob/master/strings.pdf), these are the base equivalents to stringr functions.
+
+# Detect Matches
```
hadley: Can you please use sentence case for headings?

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
...
+
+| base | stringr    |
+|------|------------|
+|   `grepl()`   | `str_detect()` |
```
hadley: Do you think it would be useful to indicate the main arguments? Like `grepl(pattern, string)` vs `str_detect(string, pattern)`?

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
...
+|   `grep()`   | `str_which()`  |
+|   `gregexpr()`   | `str_count()`  |
```
hadley: I'd do `regexpr()` -> `str_locate()` and `gregexpr()` -> `str_locate_all()`

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
...
+|  `gregexpr()`    | `str_locate()` |
+
+# Subset Strings
+
+| base | stringr    |
+|------|------------|
+|   `substr()`   | `str_sub()`    |
+|    `grep()`  | `str_subset()` |
```
hadley: I'd use `grep(value = TRUE)` just to reinforce that this is a specific use of `grep()`

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
...
+
+
+# Manage Lengths
+
+| base | stringr    |
+|------|------------|
+|   `nchar()`   | `str_length()` |
+|  `sprintf()`    | `str_pad()`    |
+|   `substr()`   | `str_trunc()`  |
+|   `trimws()`   | `str_trim()`   |
```
hadley: Should `str_wrap()` live here?

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
...
+|   `regexpr()` + `regmatches()`   | `str_extract()`|
+|   `regexec()` + `regmatches()`   | `str_match()`  |
+
+
+# Manage Lengths
+
+| base | stringr    |
+|------|------------|
+|   `nchar()`   | `str_length()` |
+|  `sprintf()`    | `str_pad()`    |
```
hadley: `sprintf()` feels a bit too general where. I think I'd remove it.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
...
+|   `substr()`   | `str_trunc()`  |
```
hadley: Whereas here `str_trunc()` does quite a bit more than `substr()`.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
...
+|  `iconv()`    | `str_conv()`     |
+|  N/A    | `str_view()`     |
+|   N/A   | `str_view_all()` |
+|  `strwrap()`    | `str_wrap()`     |
+
+```{r}
+library(stringr)
+```
+
+The following examples walk through the examples in the stringr documentation and reproduce the results in base R.
```
hadley: I think this needs a top-level heading, but I'm not entirely sure what the purpose is. Currently it seems like the goal is to allow people to replace stringr with base, which seems like a bit of a weird goal.

It's also rather long, and that along with the lack of a clear organising principle makes it a bit hard to review. So I think we should discuss here a bit and then come up with a plan for revisions.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,1202 @@
...
+str_to_lower(dog)
+str_to_title(dog)
+
+## base
+toupper(dog)
+tolower(dog)
+tools::toTitleCase(dog)
+```
+
+Locale distinctions are not made in base R.
```
hadley: They are made, they're controlled with global variables (i.e. `Sys.setlocale()`)

vignettes/compareToBaseR.Rmd
sastoudt: Thanks for the feedback! I'll go back and make the changes for all of your suggestions, and then I am happy to make a plan for larger scale revisions.

The vignette is certainly a bit unwieldy, and I can definitely cut and rearrange based on what you think makes sense. My original purpose was to show how gnarly some of the base R equivalents to simple stringr calls are, but I can see how heavy that ends up being.

I could organize the functions by the chunks in the tables above instead of in alphabetical order to make the vignette easier to navigate. Perhaps it is enough to show the difference in output of comparable functions to make it clear how un-tidy the output of base R functions can be (as opposed to going through the effort to make the outputs look identical). 

I also like your idea of emphasizing the difference in expected order of inputs between base and stringr.

hadley: Random thought: would it be better to move the individual base R vs string examples into the function documentation?

hadley: But first, maybe it's making sure we're clear on the goal. Is it to:

* help people transition from base to stringr?
* promote stringr by showing specific use cases?
* help people transition from stringr to base if they want to remove a package dependency?

I think the first is probably the most important for me.

sastoudt: Your preferred goal makes sense; that is probably the biggest use case.

How about I plan for the next draft to be organized by functionality (detect matches, subset strings, manage lengths, mutate strings, order strings, join, and split), start with the base R example (minus the extra effort to tidy the output), and then introduce the equivalent stringr function (noting the difference in input order and the differences in output formatting)? I can also just pick the most representative example(s) for each function rather than doing each example provided in the documentation. This should streamline the vignette and make it more easily navigable. 

Outlining Example:

Suppose you want to know whether each word in a vector of fruit names contains an "a". In base you would use `grepl(pattern = "a", x = fruit)`. In stringr, the pattern comes last: `str_detect(x = fruit, pattern = "a")`. [maybe make a point about how this order of inputs is helpful for `apply()` and `purrr::map()` statements?]

Then move on to `grep()` and note that it can be manipulated to do various things, but stringr compartmentalizes functionality more (i.e. `grep()` can be used for both `str_which()` and `str_subset()`).

etc...

Then maybe at the end, note some common themes in switching from base R functions to stringr functions overall.

hadley: Yeah, that sounds great!

vignettes/compareToBaseR.Rmd
sastoudt: done in commit 70d0f45

vignettes/compareToBaseR.Rmd
sastoudt: done in commit 70d0f45

vignettes/compareToBaseR.Rmd
sastoudt: done in commit 70d0f45

vignettes/compareToBaseR.Rmd
sastoudt: done in commit 70d0f45

vignettes/compareToBaseR.Rmd
sastoudt: done in commit 70d0f45

vignettes/compareToBaseR.Rmd
sastoudt: done in commit 70d0f45

vignettes/compareToBaseR.Rmd
sastoudt: done in commit 70d0f45

vignettes/compareToBaseR.Rmd
sastoudt: done in commit 70d0f45

vignettes/compareToBaseR.Rmd
sastoudt: done in commit 70d0f45

vignettes/compareToBaseR.Rmd
hadley: BTW you can just click "resolve conversation" — you don't need to say when

yutannihilation:
Travis fails with this error:

> there is no package called 'plyr'

If you need to use `plyr::rbind.fill()`, you also need to add plyr package to `Suggests`.

hadley:
Or better, eliminate (or don't run) the example that uses plyr to avoid taking another dependency (suggested dependencies aren't bad, but I don't think it's worth it for the sake of a single example)

sastoudt:
Thank you for your patience! I've shortened and refocused the vignette, and removed the `plyr` dependency. I still have some `stringr::str_wrap()` and `strwrap()` issues, and the Travis CI check fails on R version 3.1.

I'm happy to do another round of edits based on your feedback.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+```
+
+Now we will walk through each of the functions in the above tables, noting the similarities and differences between base R and stringr.
+
+These stringr examples come from the [stringr documentation](https://cran.r-project.org/web/packages/stringr/stringr.pdf) and here, they are contrasted with the analogous base R operation(s).
+
+
+# Detect matches
+
+## `str_detect`
```
hadley: Maybe:

```
## `str_detect()`: Detect the presence or absence of a pattern in a string
```

(note the parentheses after the function name; that's house style)

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+
+*Detect the presence or absence of a pattern in a string*
+
+Suppose you want to know whether each word in a vector of fruit names contains an "a". 
+
+```{r}
+fruit <- c("apple", "banana", "pear", "pinapple")
+
+## base
```
hadley: Can you please use a single `#`?

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+```{r}
+## base 
+gregexpr(pattern = "a", text = fruit, fixed = TRUE)
+
+
+## stringr
+str_count(string = fruit, pattern = "a")
+```
+
+This information can be gleaned from `gregexpr()` in base, but you will need to parse the output. This functionality is much more straightforward in stringr's `str_count()`.
```
hadley: Can you actually do the computation?

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+```{r}
+## base
+gregexpr(pattern = "p", text = fruit)
+
+## stringr
+str_locate(string = fruit, pattern = "p")
+str_locate_all(string = fruit, pattern = "p")
+```
+
+Again we would need to parse the output of `gregexpr()` to get the information we want. The stringr package provides specialized functions, `str_locate()` and `str_locate_all()` to do the job.
```
hadley: I think you should also comment on what happens when there isn't a match.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+
+## stringr
+str_sub(string = hw, start = 1, end = 6)
+str_sub(string = hw, start = 1)
+str_sub(string = hw, end = 6)
+```
+
+In base you could use `substr()` or `substring()`. The former requires both a start and stop of the substring while the latter assumes the stop will be the end of the string. The stringr version, `str_sub()` has the same functionality, but also gives a default start value (the beginning of the string). Both the base and stringr functions have the same order of expected inputs. 
+
+
```
hadley: I think you should next talk about the use of negative numbers with stringr, and how the functions are vectorised.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+There are some differences in behavior between `str_sub()` and `substr()`.
```
hadley: I think this should be in a separate section, since this is technically a different function `str_sub<-` — I'd call this subset assignment or similar.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+```
+
+If we want to extract matches in base R we need to use the output of `regexpr()` as input to `regmatches()` whereas in stringr we would only  need one function call to `str_extract()`. If we want to extract all matches instead of just the first, we would need to wrangle the output of base R's `gregexpr()`. In stringr, there is a specialized function to do this: `str_extract_all()`.
+
+
+## `str_match`
+
+*Extract matched groups from a string*
+
+We may also want to extract groups from a string. Here I'm going to use the scenario from Section 14.4.3 in [R for Data Science](https://r4ds.had.co.nz/strings.html) instead of the one in the documentation because I find the nouns in sentences example easier to conceptualize than the phone number example.
```
hadley: I don't think you need to explain this decision

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+do.call("rbind", regmatches(x = has_noun, m = matches))
+gregexpr(pattern = noun, text = has_noun)
+
+## stringr
+str_match(string = has_noun, pattern = noun)
+str_match_all(string = has_noun, pattern = noun)
+
+```
+
+The output of grouped matches is the full match and then the components of each group. Again, in base R we need to pair `regexec()` and `regmatches()` whereas in stringr we just need one function: `str_match()`. To get all the grouped matches in base R we would need to organize the output of `gregexpr()` whereas stringr's `str_match_all()` streamlines this process.
```
hadley: I have a vague recollection that there's some complication with the base approach in the absence of a match.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+
+## stringr
+str_length(letters)
+
+```
+
+There are some subtle differences between base and stringr here. `nchar()` requires a character vector, so it will return an error if used on a factor. `str_length()` can handle a factor input.
+
+
+```{r,eval=F}
```
hadley: Better to use `error = TRUE` rather than `eval = FALSE` (then you can remove the comment recreating the error message)

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+```
+
+## `str_pad`
+
+*Pad a string*
+
+If you want to pad a string to a certain width, you can use stringr's `str_pad()`. In base R you could use `sprintf()`, but unlike `str_pad()`, `sprintf()` has many other functionalities. 
+
+```{r}
+## stringr
```
hadley: I think you should consistently put the base R code first

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+  str_trunc(x, 20, "center")
+)
+
+```
+
+## `str_trim`
+
+*Trim whitespace from a string*
+
+Similarly, stringr provides `str_trim()` to trim whitespace from a string. This is analogous to base R's `trimws()`.
```
hadley: Might be worth mentioning which version of R `trimws()` was added. IIRC it was fairly recently.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+## stringr
+str_squish(" String with trailing, middle,   and leading white space\t")
+str_squish("\n\nString with excess, trailing and leading white space\n\n")
+```
+
+## `str_wrap`
+
+*Wrap strings into nicely formatted paragraphs*
+
+I'm not sure why these outputs don't match. The stringr packages's `str_wrap()` is a wrapper around `stringi::stri_wrap()` which has defaults that are supposed to be compatible with base R's `strwrap()`.
```
hadley: I don't think the implementations are supposed to match — stringi uses Knuth-Plauger

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+
+*Replace matched patterns in a string*
+
+To replace certain patterns within a string, stringr provides the functions `str_replace()` and `str_replace_all()`. The base R equivalents are `sub()` and `gsub()`. Note the difference in default input order again. In stringr the string comes first, followed by the pattern and replacement, while in base R, the pattern and replacment come before the string.
+
+```{r}
+fruits <- c("apple", "banana", "pear", "pinapple")
+
+## stringr
+str_replace(string = fruits, pattern = "[aeiou]", replacement = "-")
```
hadley: I wouldn't name the arguments here

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+str_to_lower(dog)
+str_to_title(dog)
+
+## base
+toupper(dog)
+tolower(dog)
+tools::toTitleCase(dog)
+```
+
+In stringr we can control the locale, while in base R locale distinctions are controlled with global variables.
```
hadley: I think you can eliminate the base R example, but you should emphasise that this means that the same base R code might return different results when run on different computers.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+*Join multiple strings into a single string*
+
+To join multiple strings into one, base R uses `paste()` while string_r uses `str_c()` (the "c" stands for concatenate).
+
+```{r}
+## stringr
+str_c("Letter: ", letters)
+
+## base
+paste("Letter: ", letters, sep = "")
```
hadley: `paste0()` is a fairer comparison here. I don't think there's much benefit to using `str_c` these days.

It might be worth mentioning `str_flatten()` as a function always returns a character vector of length 1.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+
+## stringr
+str_glue(
+  "My name is {name}, ",
+  "my age next year is {age + 1}, ",
+  "and my anniversary is {format(anniversary, '%A, %B %d, %Y')}."
+)
+
+## base
+paste("My name is ", eval(name), ", my age next year is ", eval(age + 1), ", and my anniversary is ", eval(format(anniversary, "%A, %B %d, %Y")), ".", sep = "")
```
hadley: No need for `eval` here

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+## stringr
+str_order(letters)
+str_sort(letters)
+
+## base
+order(letters)
+sort(letters)
+```
+
+Some options in `str_order()` and `str_sort()` don't have analogous base R options.
```
hadley: Again, I don't think you need an example, but you should emphasise that the base functions can return different results for the same inputs.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+```{r}
+x <- c("100a10", "100a5", "2b", "2a")
+str_sort(x)
+str_sort(x, numeric = TRUE)
+
+sort(x)
+## no numeric capability in sort
+```
+
+# Helpers
```
hadley: I think you can drop this section

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+
+The stringr package provides a nice way to look at regular expression matches to help with text processing. 
+
+```{r}
+## stringr
+str_view(c("abc", "def", "fgh"), "[aeiou]")
+```
+
+
+# Take-Aways
```
hadley: I think you should move this to before detect matches — that way you can discuss argument order in one place.

vignettes/compareToBaseR.Rmd
```diff
@@ -0,0 +1,683 @@
...
+
+Overall the main differences between base R and stringr are:
+
+1. The stringr package uses the `str_` prefix for its functions while base R has no unifying format for its functions.
+2. The order of inputs is often (but not always) opposite between base R and stringr. In base R, the pattern often comes before the string while in stringr the expected input order is flipped. The stringr order of inputs can be useful when using `apply()` or `purrr::map()` statements.
```
hadley: I wouldn't say "opposite"; I'd say different, and emphasise that it's totally consistent for stringr functions.

hadley:
Thanks Sara, I'll take it from here 😄

hadley:
Thanks @sastoudt — I made a few more tweaks to make it a bit more concise, and added a few more details where appropriate. I really appreciate all your work on this!

hadley:
Let me know if there's anything I can do to help you to get academic credit for this — if you remind me in a few months, I can give you an estimate of how many people have read it. The stringr site is the 6th most popular in the tidyverse, so it should be read by at least a few thousand people,

sastoudt:
Awesome, thank you! I'll definitely check in for some stats in a few months for my CV.

PR #289, opened 2019-03-04
add str_split_n(); closes #278

### LINKED ISSUE #278: str_split_n(string, pattern, n)
Returns a character vector.

For the moderately common case where you only want to extract a single piece from a string.

Inspired by `SPLIT_PART()` from SQL.

bfgray3:


R/split.r
```diff
@@ -50,3 +58,10 @@ str_split_fixed <- function(string, pattern, n) {
   out[is.na(out)] <- ""
   out
 }
+
+#' @export
+#' @rdname str_split
+str_split_n <- function(string, pattern, n) {
+  out <- str_split(string, pattern, simplify = TRUE)
+  apply(out, 1, `[`, i = n)
```
hadley: I don't think `apply()` is the right function here because `str_split()` returns a list, not a matrix. Could you please use `vapply()` + `[[` instead?

It'd also be worth thinking about what happens when there are fewer pieces than `n`? Should it be an error? Or return `NA`? Or `""`?

bfgray3: Sorry about the `apply()`--I originally was using `simplify = TRUE` in `str_split()` which I now realize wasn't a great idea.  I just pushed a commit that returns `NA` when there are fewer pieces than `n` but I'd like to get your thoughts about what you think might be most natural to do here.

hadley:
Thanks! I think `NA` is the right thing to do here.

PR #297, opened 2019-05-15
Fix typo in "from base" vignette

michaelweylandt:
Second argument to `str_dup` should match second
argument to `strrep` (both are `n` - number of times
to repeat).


hadley:
Thanks!

Issue #301, opened 2019-06-04
test for test-pad has glitch?


richardjtelford:
The test for in test-pad for "long strings are unchanged" has the expectation 

```r
expect_equal(str_length(padded), str_length(padded))
```

This will always be true.

Presumably it should be

```r
expect_equal(str_length(padded), str_length(strings))
```

PR #302, opened 2019-06-05
fixing expectation #301

richardjtelford:
changing expectation for test-pad so that it can fail


hadley:
Thanks!

For future reference, if you include `Fixes #xyz` in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

PR #308, opened 2019-08-13
Add str_replace() example to help text

IndigoJay:
Add example to help file on multiple pattern matching. Discussed at https://community.rstudio.com/t/understanding-the-use-of-str-replace-all-with-multiple-patterns/1849/2


hadley:
That didn't feel like quite the right place to explain this idea — I think a vignette would make more sense because it's a principle that applies to all functions. But I did add a comment to explain those examples which were a bit confusing.

PR #310, opened 2019-09-14
Add str_left(), str_right() and str_mid() functions.

chrimaho:
Requesting to add the `str_left()`, `str_right()` and `str_mid()` functions to the `sub.r` file. These new functions are simply wrappers around the `base::substr()` function. The reason for this request is to make the syntax and function naming similar to the Excel functions that to the same process.  The whole purpose of these functions is to make the life of an R Coder easier; especially if they are newly upgrading from the Excel environment. Essentially, these functions are direct equivalents to the very popular Excel functions: `LEFT()`, `RIGHT()` and `MID()`.

### DIFF
--- a/R/sub.r
+++ b/R/sub.r
@@ -80,3 +80,130 @@ str_sub <- function(string, start = 1L, end = -1L) {
   }
   string
 }
+
+
+
+#' @title
+#' Select the LEFT, RIGHT or MID of a string.
+#'
+#' @description
+#' These functions are identical to Excel equivalents of `LEFT()`, `RIGHT()` and `MID()`.
+#' All three functions (`str_left()`, `str_right()` and `str_mid()`) are wrappers around `base::substr()`.
+#' The concept and Examples are provided by John (https://www.rforexcelusers.com/author/jtav04gmail-com/) on
+#'     R For Excel Users (https://www.rforexcelusers.com/how-to-mid-right-left-r/).
+#'
+#' @details
+#' Vectorised over `string` to length `num_chars`, starting from `start_num`.
+#'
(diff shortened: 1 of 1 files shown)


cderv:
I find this a nice idea to have some excel-like function like that. It could help transitions. 👍 
Just a tip I wanted to share: `stringr` is already heavily based on `stringi` and you have a `stringi::stri_sub()` that exists and can be useful alternative to base function to get more closely aligned with other functions.

chrimaho:
Hey mate, Thanks!
I definitely agree that it would be better to align to `stringi` functions than to `base` functions. And when I was reviewing the `stringr::str_sub()` function, I saw that it also referenced `@seealso The underlying implementation in [stringi::stri_sub()]`.
I included those `base` examples in my code as a point of reference so that people can see that there were already other ways of achieving this functionality. So, I'm not really re-inventing the wheel here, just re-phrasing it to make the Excel-to-R transition easier for our future R community.
Would you like me to change the examples to reference the `stringi` functions, then re-submit the Pull Request?

cderv:
No. Don't change anything. I just wanted to give cheers to this idea and share some thoughts. 
I used some wrappers once using `stringr` function instead of based R to get it vectorized.
``` r
str_left <- function(string, num_chars) {
  stringr::str_sub(string, start = 1, end = num_chars)
}
str_left("abcdfe", c(1, 2))
#> [1] "a"  "ab"
```
That's why I wanted to give a big +1 to the idea. 😄

hadley:
I appreciate your enthusiasm but the code style is very far away from the existing stringr style. I'm also not terribly convinced by the utility of these functions given the flexibility of `str_sub()`, and I'm worried that stringr is getting too many functions that do very similar things.

PR #315, opened 2019-11-05
Implement str_like() closes #280

### LINKED ISSUE #280: Implement str_like()
Which follows the SQL `LIKE` definition

rjpat:
Created a str_like() function to match the SQL operator's wild cards and escaping. Works by converting the SQL version to their regex equivalents.
Closes #280


rjpat:
Split the conversion into its own function and test set.

R/detect.r
```diff
@@ -104,3 +104,53 @@ str_ends <- function(string, pattern, negate = FALSE) {
...
+#'
+#' Turns a SQL LIKE structure into a recognisible regular expression pattern.
+#'
+#' @param pattern SQL LIKE operator pattern to be converted.
+#'
+#' @return A converted pattern.
+#'
+#' @examples
+#' like_converter("bana%")
+#' like_converter("app_e")
```
hadley: You don't need to document this function because it's not exported; and generally I think it's better to put little helper functions like this at the bottom of the file (so you see the big picture first, before the details)

R/detect.r
```diff
@@ -104,3 +104,53 @@ str_ends <- function(string, pattern, negate = FALSE) {
...
+#' @return A logical vector.
+#' @export
+#'
+#' @examples
+#' fruit <- c("apple", "banana", "pear", "pinapple")
+#' str_like(fruit, "app%")
+#' str_like(fruit, "ba_ana")
+#' str_like(fruit, "%APPLE")
+str_like <- function(string, pattern, negate = FALSE, ignore_case = TRUE) {
+  switch(type(pattern),
```
hadley: I think you could now simplify this to:

```
if (type(pattern) == "regex") {
  pattern <- like_converter(pattern)
}
str_detect(string, pattern, negate = negate, ignore_case = ignore_case)
```

Does that make sense?

hadley: Hmmm, or should this function accept the other pattern types at all? Maybe that doesn't make sense for a more specialised function. In that case the function would look more like:

```R
switch(type(pattern), 
  regex = stri_detect_regex(string, like_converter(pattern), negate = negate, opts_regex = opts(pattern)),
  stop("str_like() works only with regular expressions", call. = FALSE
)
```

R/detect.r
hadley: And I think `like_to_regex` would be a slightly better name.

tests/testthat/test-detect.r
```diff
@@ -49,3 +49,26 @@ test_that("str_ends works", {
...
+  expect_equal(like_converter("ab%"), "^ab.*$")
+  expect_equal(like_converter("ab_"), "^ab.$")
+
+  # escaping
+  expect_equal(like_converter("ab\\%"), "^ab\\%$")
+  expect_equal(like_converter("ab[%]"), "^ab[%]$")
+})
+
+test_that("str_like works", {
+  expect_true(str_like("abc", "a_c"))
```
hadley: Once you've rewritten `str_like()` as I suggested above you'll be able to substantially simplify this test.

R/detect.r
rjpat: I agree that this should probably just accept a regex, have updated similar to above and made a separate call in order to preserve the ignore case earlier.

tests/testthat/test-detect.r
rjpat: Have kept only one of the general test for `str_like()` as we are testing the `like_to_regex` converter earlier. Do you think there is any further way to simplify/should there be a test to ensure it fails on any other type of regex now?

hadley: I decided there was a slightly better way to check `pattern` so I tweaked the test too

hadley:
Thanks for all your hard work on this!

PR #323, opened 2020-01-11
Remove periods from end of doc titles

dbarrows:
According to [The tidyverse style guide - 7.2 Title and description](https://style.tidyverse.org/documentation.html#title-and-description):

> "Titles should use sentence case but not end with a full stop (.)."

This PR removes periods from the ends of the documentation titles that have them.


hadley:
Thanks @dbarrows — this is great!

PR #328, opened 2020-02-06
Fix str_starts() and str_ends() with negate

clemenshug:
This is fixing issue #327 for me. I'm not quite sure it's OK to assume `negate` is scalar, what do you think?

### DIFF
--- a/tests/testthat/test-detect.r
+++ b/tests/testthat/test-detect.r
@@ -28,10 +28,18 @@ test_that("modifiers work", {
 test_that("str_starts works", {
   expect_true(str_starts("ab", "a"))
   expect_false(str_starts("ab", "b"))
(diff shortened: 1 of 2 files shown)


hadley:
Fixes #327

tests/testthat/test-detect.r
```diff
@@ -28,10 +28,18 @@ test_that("modifiers work", {
 test_that("str_starts works", {
   expect_true(str_starts("ab", "a"))
   expect_false(str_starts("ab", "b"))
+  expect_true(str_starts("ab", fixed("a")))
```
hadley, line 31: I think you should only need 4 new tests — one each for `str_starts()` and `str_end()`, and `fixed` and `coll`

PR #340, opened 2020-05-30
make detect helpers honor regex "|" operator

carlganz:
Apropos of this [twitter thread](https://twitter.com/carlishganzino/status/1266519160374190081?s=20) here is small PR that updates detect helpers so that they honor regex "|" operator more intuitively. It doesn't break any existing tests, but I didn't run full revdepcheck because I don't want my computer to catch fire. 

I'm sure there will be cases where somebody has code that depends on the previous functionality, but I feel pretty strongly that this change makes the API more intuitive.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # stringr (development version)
 
+* Update `str_starts()` and `str_ends()` functions so they honor regex "|" operator more intuitively. (@carlganz)
+
 * `word()` now returns all the sentence when using a negative `start` parameter
   that is greater or equal than the number of words. (@pdelboca, #245)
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # stringr (development version)
 
+* Update `str_starts()` and `str_ends()` functions so they honor regex "|" operator more intuitively. (@carlganz)
```
hadley, line 3: I’d phrase this more generally - it’s about operator precedence in generally, not just `|`. Otherwise, looks great.

tests/testthat/test-detect.r
```diff
@@ -36,6 +36,10 @@ test_that("str_starts works", {
   # Special typing of patterns.
   expect_true(str_starts("ab", fixed("A", ignore_case = TRUE)))
   expect_true(str_starts("ab", regex("A", ignore_case = TRUE)))
+
+  # Or operators are respected
+  expect_true(str_starts("ab", regex("B|A", ignore_case = TRUE)))
```
hadley, line 41: Also I don’t think you should using `ignore_case` as that’s unrelated to this isse

hadley:
Thanks!

Issue #195, opened 2017-11-16
str_split_fixed should return NA not ""


hadley:
``` r
reprex::reprex_info()
#> Created by the reprex package v0.1.1.9000 on 2017-11-16

library(stringr)
str_split_fixed(fruit[31:33], pattern = " ", n = 2)
#>      [,1]         [,2]   
#> [1,] "fig"        ""     
#> [2,] "goji"       "berry"
#> [3,] "gooseberry" ""
```

To be more consistent with other stringr function.

hadley:
Will need to wait for major release

PR #367, opened 2021-03-11
return all NA rows/cols for no matches in matrix results

### LINKED ISSUE #366: return matrices of all NA rows/cols for all no-matches
`str_extract_all(simplify = TRUE)` currently returns what might be considered inconsistently, depending on whether there is at least one match:

``` r
# one match: returns matrix with empty string for no matches
stringr::str_extract_all(c("foo", "bar"), pattern = "bar", simplify = TRUE)
#>      [,1] 
#> [1,] ""   
#> [2,] "bar"
# no matches: returns matrix with 0 cols
stringr::str_extract_all("foo", pattern = "bar", simplify = TRUE)
#>     
#> [1,]
stringr::str_extract_all(c("foo", "zap"), pattern = "bar", simplify = TRUE)
#>     
#> [1,]
#> [2,]
```

<sup>Created on 2021-03-10 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

I'd imagine that this is simply because of base R, which will drop empty (character) vectors.
However, stringr already seems to be "repair" this base R behavior when there are at least *some* matches:

``` r
# 0-length char just gets dropped
c("foo", character())
#> [1] "foo"
# resulting 0-length char gets changed to ""
stringr::str_extract_all(c("foo", "bar"), pattern = "bar", simplify = TRUE)
#>      [,1] 
#> [1,] ""   
#> [2,] "bar"
```

The [docs](https://stringr.tidyverse.org/articles/from-base.html) also state:

> Base R requires the combination of `regexpr()` with `regmatches()`; but note that the strings without matches are dropped from the output. stringr provides `str_extract()` and `str_extract_all()`, and the output is always the same length as the input.

It might be great if stringr could always hide away this base R behavior and always return at least a 1x1 matrix.
Otherwise, whatever function is calling the result and need to take special precaution for this special case (`x[, 1]` might fail, for example).
I'm also not sure how whether this behavior could be considered length-stable.
(It *is* type-stable, because the result *is* still a matrix).

maxheld83:
Closes #366.

The matrix returns from `str_extract_all(simplify = TRUE)`, `str_locate_all()` and `str_match_all()` would previously drop rows or columns without any matches (details and reprexes in #366).
This can make it harder to work with these functions, because callers have to anticipate differently structured results (for example `x[1, ]`) may fail).
The old behavior made these functions length-unstable (am I using this right?).

If this is even a desirable change, the biggest problem I can see is that this is a (small) **breaking change**.
I'll address the sources of breaking separately:

1. Some other package code might now expect the above dropping behavior.
    Happily, this dropping was nowhere documented nor tested.
    Some `ncol(x) == 0` or similar conditions (if they are ever used) would no longer trigger, but presumably however reverse dependent code dealt with `NA`s would still work, and may cover an *all*-`NA` row/col just fine.
    This is speculation and will require someone with more experience and maybe revdepcheck results.
2. As a corollary of these fixes, `str_extract_all(simplify = TRUE)` now returns `NA`s for no matches, not `""`.
     I'm guessing this is more likely to be a breaking change.
     It is also not currently documented or tested, but still.
     One *could* work around this with by replacing `NA`s with `""` again, though the long-term plan (as per #195) seems to be to standardise on `NA` anyway, so this might be a stop-gap and cause headaches twice.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # stringr (development version)
 
+* Always return at least 1x1 matrix from `str_extract_all(simplify = TRUE)`.
+  This introduces a small, potentially breaking change:
+  No matches are now returned as `NA_character`, not as `""`.
+  (@maxheld83, #366)
+
 * Update `str_starts()` and `str_ends()` functions so they honor regex operator precedence. (@carlganz)
 
 * `word()` now returns all the sentence when using a negative `start` parameter
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,9 +1,12 @@
 # stringr (development version)
 
-* Always return at least 1x1 matrix from `str_extract_all(simplify = TRUE)`.
+* `str_extract_all(simplify = TRUE)` and `str_match_all()` now always return
+  all combinations of given `string` and `pattern` as a matrix of their 
+  cartesian product.
+  Previously, rows or columns without matches would be dropped.
   This introduces a small, potentially breaking change:
-  No matches are now returned as `NA_character`, not as `""`.
-  (@maxheld83, #366)
+  No matches in matrices are now returned as `NA_character` in
+  `str_extract_all(simplify = TRUE)`, not `""`. (@maxheld83, #366)
 
 * Update `str_starts()` and `str_ends()` functions so they honor regex operator precedence. (@carlganz)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,12 +1,12 @@
 # stringr (development version)
 
-* `str_extract_all(simplify = TRUE)` and `str_match_all()` now always return
-  all combinations of given `string` and `pattern` as a matrix of their 
-  cartesian product.
-  Previously, rows or columns without matches would be dropped.
-  This introduces a small, potentially breaking change:
-  No matches in matrices are now returned as `NA_character` in
-  `str_extract_all(simplify = TRUE)`, not `""`. (@maxheld83, #366)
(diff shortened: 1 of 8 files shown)


maxheld83:
`str_extract_all(simplify = TRUE)` has another wrinkle, courtesy of `stringi::stri_extract_all()`:

``` r
stringi::stri_extract_all(c("a b b", "a"), regex = "b", simplify = TRUE)
#>      [,1] [,2]
#> [1,] "b"  "b" 
#> [2,] NA   ""
```

So when there's not a match, it's `NA`, but when there's nothing to match *against* it's `character()`.
There's a difference here, but I think the resulting `NA` *and* `""` missing can be confusing.

`stringi::stri_extract_all()` *does* have offer the desired behavior via `simplify = NA`:

``` r
stringi::stri_extract_all(c("a b b", "a"), regex = "b", simplify = NA)
#>      [,1] [,2]
#> [1,] "b"  "b" 
#> [2,] NA   NA
```

This is now https://github.com/tidyverse/stringr/pull/367/commits/947a20f83c84c614193b8a0526a3cdacb29b3269 the behavior in stringr: it treats `TRUE` as `NA`.
Deviating from stringi behavior for a shared argument is pretty bad, but it seems like the least bad solution here.

hadley:
Doesn't this violate the invariant that the number of rows equals the number of matches?

hadley:
Yeah, lets discuss more in that issue.

PR #376, opened 2021-04-22
Implement tidyverse recycling rules

### LINKED ISSUE #372: Implement tidyverse recycling rules
i.e. only recycle vectors of length 1

hadley:
Will fix #372 when complete.


hadley:
Re-thinking `str_c()` recycling for performance — can maybe just `compact()`? Need to test all combinations of expected behaviour (with `character()` and `NULL`)

PR #391, opened 2021-07-02
Implement `stringr_locale()`

DavisVaughan:
This PR adds `stringr_locale()`, a new helper for controlling the default locale used by stringr in any function that has a `locale` argument. It has the following properties:

- Defaults to `"en"` to retain backwards compatibility
- The default is customizable by setting `tidyverse.locale_collation`
- Remaps `"C"` to `"en_US_POSIX"` in case a user set the locale to `"C"` when using dplyr


R/locale.R
```diff
@@ -0,0 +1,40 @@
...
+#' @export
+#' @examples
+#' stringr_locale()
+#'
+#' withr::with_options(
+#'   list(tidyverse.locale_collation = "fr"),
+#'   stringr_locale()
+#' )
+stringr_locale <- function() {
+  locale <- peek_option("tidyverse.locale_collation")
```
DavisVaughan, line 23: I just want to gut check this one more time. Are we sure that we want this to be `tidyverse.locale_collation` which is shared between dplyr and stringr? Would it be surprising for users who set this argument specifically while working on dplyr to then find that stringr is returning different results?

Something about entangling the two packages together through this option feels _slightly_ wrong to me. Like, I feel like there should be `stringr.locale` and `dplyr.locale`. Setting each one of those individually doesn't feel like too much work, and then you avoid the above mentioned confusion and you don't have to remap C->en_US_POSIX.

hadley: Feels like it's worth collectively discussing this somewhere? (Maybe just in slack). It feels to me that we really want a unified notion of tidyverse locale, not one local per package.

DavisVaughan:
We've collectively decided not to implement this. Rather than having a tidyverse wide locale global option, we are instead going to restrict it to just a "convenience option" for dplyr `arrange()`. Having global options that affect computations is not best practice, so we'd like to minimize their usage where possible.

PR #422, opened 2022-01-20
Fix knitr installation on r-devel

jennybc:
dev glue has a fix for the problem cropping up in stringr's CI where knitr fails to install

More detail is here: https://github.com/tidyverse/glue/pull/254

stringr is basically a bystander here, though stringr's `importFrom(glue,glue)` is part of the story


hadley:
Thanks!

PR #462, opened 2022-10-04
Vectorise replacement function

hadley:
This breaks 7 revdeps but I think it's worth it because the performance improvement is immense.


hadley:
It also breaks rmarkdown 😬 

```
Error in `if (grepl("^[.][.]", in_file)) ...`:
! the condition has length > 1
---
Backtrace:
     ▆
  1. └─pkgdown::build_article("from-base", new_process = FALSE)
  2.   └─pkgdown:::render_rmarkdown(...) at pkgdown/R/build-articles.R:281:2
  3.     ├─rlang::inject(rmarkdown::render(!!!args)) at pkgdown/R/rmarkdown.R:52:4
  4.     └─rmarkdown::render(...) at rlang/R/eval.R:195:2
  5.       └─output_format$post_processor(...)
  6.         └─rmarkdown (local) base(metadata, input_file, output_file, ...)
  7.           └─rmarkdown:::process_images(output_str, image_relative)
  8.             └─rmarkdown:::process_html_res(...)
  9.               └─stringr::str_replace_all(html, reg, process_img_src)
 10.                 └─stringr:::str_transform_all(string, pattern, replacement) at stringr/R/replace.r:91:4
 11.                   └─rmarkdown (local) replacement(old_flat) at stringr/R/replace.r:201:4
 12.                     └─rmarkdown (local) processor(img_src, src)
```

hadley:
rmarkdown patch in https://github.com/rstudio/rmarkdown/pull/2416. Breaking rmarkdown is a big deal, so I don't think we should merge this until after rmarkdown is released, so it probably won't make it for this release.

DavisVaughan:
I guess pkgdown is failing because rmarkdown is broken? Very meta

R/replace.R
```diff
@@ -179,18 +179,50 @@ str_replace_na <- function(string, replacement = "NA") {
...
+  new <- vctrs::vec_chop(new_flat, idx)
+
+  stringi::stri_sub_all(string, locs) <- new
   string
 }
+
+chop_index <- function(x) {
+  ls <- lengths(x)
+  start <- cumsum(c(1, ls[-length(ls)]))
+  end <- start + ls - 1
```
DavisVaughan, line 226:
```suggestion
  start <- cumsum(c(1L, ls[-length(ls)]))
  end <- start + ls - 1L
```

R/replace.R
```diff
@@ -179,18 +179,50 @@ str_replace_na <- function(string, replacement = "NA") {
...
-    for (j in rev(seq_len(nrow(locs[[i]])))) {
-      loc <- locs[[i]]
-      str_sub(string[[i]], loc[j, 1], loc[j, 2]) <- replacement(str_sub(string[[i]], loc[j, 1], loc[j, 2]))
-    }
+  old <- str_sub_all(string, locs)
+  idx <- chop_index(old)
+
+  # unchop list into a vector, apply replacement(), and then rechop back into
+  # a list
+  old_flat <- vctrs::vec_unchop(old, idx)
```
DavisVaughan, line 195: Depending on how much you value readability over performance, you don't actually need to pass `idx` here since you are unchopping in the order the input was provided

R/replace.R
```diff
@@ -179,18 +179,50 @@ str_replace_na <- function(string, replacement = "NA") {
...
+  if (length(old_flat) == 0) {
+    # minor optimisation to avoid problems with the many replacement
+    # functions that use paste
+    new_flat <- character()
```
DavisVaughan, line 199: 💀

R/view.R
```diff
@@ -107,9 +107,13 @@ str_view_filter <- function(x, pattern, match) {
 
 str_view_highlighter <- function(html = TRUE) {
   if (html) {
-    function(x) paste0("<span class='match'>", x, "</span>")
+    function(x) str_c("<span class='match'>", x, "</span>")
   } else {
     function(x) {
+      if (length(x) == 0) {
+        return(character())
+      }
```
DavisVaughan, line 115: I guess this is where you happened to notice the cli issue 😆

tests/testthat/_snaps/replace.md
```diff
@@ -29,3 +29,16 @@
...
+    Code
+      str_replace_all("x", "x", ~1)
+    Condition
+      Error in `str_replace_all()`:
+      ! `replacement()` must return a character vector, not a number.
+    Code
+      str_replace_all("x", "x", ~ c("a", "b"))
+    Condition
+      Error in `str_replace_all()`:
+      ! `replacement()` must return a vector the same length as the input (1), not length 2.
```
DavisVaughan, line 43: My brain kind of broke on this because `replacement` is really the arg name, but i see what you are trying to do by adding `()` since it is used as a function call. I don't mind either way, just wanted to say it took me a second to realize it wasn't a typo

R/replace.R
DavisVaughan: Alternate idea is to actually apply `replacement()` and then check if `length(new_flat) == 1 && length(old_flat) == 0`. In that special case you could issue a warning (possibly even mentioning `paste0()`) and "fix" it for them.

DavisVaughan (approved):
I guess the broken packages are like rmarkdown? They didn't expect multiple strings at once?

R/replace.R
hadley: Yeah, we could do that, I just don't think it's worth the effort since we do know what the correct answer is in the 0 element case.

tests/testthat/_snaps/replace.md
hadley: I'll tweak this to make it clearer.

hadley:
Yeah, the failures seem to be cases that expected a single string. So there's always a trivial fix, which is to vectorise with `vapply()` or similar. When I do the revdep fixes, I'll try and vectorise properly so they can benefit from the performance boost.

hadley:
Fixed in rmarkdown 2.17, but I think this will have to wait until the next release since otherwise installing stringr will break older (i.e. most) rmarkdown installed.

R/replace.R
hadley: In dev vctrs:

```R
xs <- list("a", c("b", "c"), c("d", "e", "f"))

x <- list_unchop(xs)
x <- toupper(x)
x
#> [1] "A" "B" "C" "D" "E" "F"

sizes <- list_sizes(xs)
vec_partition(x, sizes)
```

hadley:
I think enough time has elapsed that we can now merge this, fixing the 7 CRAN failures with PRs.

R/replace.R
```diff
@@ -180,18 +182,50 @@ str_replace_na <- function(string, replacement = "NA") {
...
+
+  # unchop list into a vector, apply replacement(), and then rechop back into
+  # a list
+  old_flat <- vctrs::list_unchop(old)
+  if (length(old_flat) == 0) {
+    # minor optimisation to avoid problems with the many replacement
+    # functions that use paste
+    new_flat <- character()
+  } else {
+    new_flat <- replacement(old_flat)
```
hadley, line 203: @DavisVaughan do you think it's worth wrapping this to provide a more informative error? (something like `replacement` function must be vectorised?)

DavisVaughan: Are you talking about removing the `length(old_flat) == 0` special case handling?

That was what I mentioned back here, I think 
https://github.com/tidyverse/stringr/pull/462#discussion_r987127244

We went through a similar thing with `dplyr::rename_with()` where we made it so that `.fn` has to return a vector of length equal to the input length (i.e. typical size stability, like here).

For dplyr, we did get a few questions about this, so we added a `paste0(recycle0 = TRUE)` example:
https://github.com/tidyverse/dplyr/pull/6690

I think you could do a few things to mitigate issues:
- Add an example of `paste0`, like dplyr
- You already have a error below when `length(new_flat) != length(old_flat)`, but you could specially call out that `replacement` needs to be _vectorized_ there as well.
- If you're feeling particularly friendly, you could check to see if `replacement` was identical to `paste` or `paste0` in the check below after `length(new_flat) != length(old_flat)` fails. If so, you could throw a special error message just for that very common case, and probably advise to use `recycle0`.

hadley: If you look at the revdeps, a lot of the errors are folks passing vectors to functions that expect scalars. So I was thinking it might make sense to wrap the call to `replacement()` to provide some hint in the error message.

DavisVaughan: Ah like

```r
    Error in if (x > 0) "+" else "-" : the condition has length > 1
    Calls: <Anonymous> ... <Anonymous> -> str_transform_all -> replacement -> <Anonymous>
```

Ok now I understand what you mean.

I think it would be quite hard to make that error precise, i.e. you can't tell when they've provided a function that only expects a scalar very easily.

So the alternative is to _always_ wrap any error from `replacement()` with some general advice that talks needing a vectorized function. I think that could be useful if you frame it like:

```
Error in str_fn():
Failed to apply `replacement()`. Check the following in addition to the error thrown by `replacement()`:
- `replacement()` must be a vectorized function capable of handling input of any length.
- `replacement()` must return a character vector with output the same length as the input.
Caused by error in replacement():
<parent error here>
```

hadley: Yeah, exactly!

PR #470, opened 2022-10-20
Reword the description and a `@param`

jennybc:


R/detect.R
```diff
@@ -1,15 +1,16 @@
 #' Detect the presence/absence of a match
 #'
-#' `str_detect()` returns a logical vector `TRUE` if `pattern` is found within
-#' each element of `string` or a `FALSE` if not. It's equivalent
+#' `str_detect()` returns a logical vector with `TRUE` for each element of
+#' `string` that matches `pattern` and `FALSE` otherwise. It's equivalent to
```
jennybc, line 4: This is the change that motivates the PR. The original is missing a few words or punctuation? So I took a stab at rewording the description to deal with both.

I got here because this documentation is also inlined into the `man.Rmd` chapter of R Packages, so it needs some kind grammatical fix. Either mine or something else.

R/detect.R
```diff
@@ -1,15 +1,16 @@
...
 #'
 #' @param string Input vector. Either a character vector, or something
 #'  coercible to one.
 #' @param pattern Pattern to look for.
 #'
-#'   The default interpretation is a regular expression, as described
-#'   `vignette("regular-expressions")`. Control options with [regex()].
+#'   The default interpretation is a regular expression, as described in
+#'   `vignette("regular-expressions")`. Use [regex()] for finer control of the
+#'   matching behaviour.
```
jennybc, line 13: This is a drive-by suggestion that I'm making now that I'm here. But it is the reason that multiple `.Rd` files are affected, due to `pattern`'s docs being inherited in a few places.

hadley:
Thanks!

PR #482, opened 2022-12-05
Skip tests where htmltools+htmlwidgets are required

MichaelChirico:


hadley:
This made realise that the tests weren't very well organised, so I moved the tests for the legacy HTML mode into one place.

PR #483, opened 2022-12-06
Tweak "From base R" vignette:

salim-b:
I think the new [*base R <-> stringr* comparison table](https://stringr.tidyverse.org/articles/from-base.html#overall-differences) is incredibly useful! Just want to improve it in the following ways:

- Define data in tribble and generate MD table from it.
- The new table is ordered alphabetically by *stringr function name* for now, which I think is more accessible (base R API is just a mess whether we sort it alphabetically by name or not). But order can easily be changed later on if desired.
- ~~The new table has links to stringr's pkgdown reference when built by pkgdown thanks to `downlit::autolink_url()`.~~ *Removed as per @hadley's request.*
- Add missing `str_extract_all()` base R equivalent
- Fixed fn param names in table.
- Fix typo in fn name.
- Use comment-style in-body YAML chunk opts
- Cosmetic changes


.Rbuildignore
```diff
@@ -11,7 +11,9 @@
 ^codecov\.yml$
 ^\.httr-oauth$
 ^_pkgdown\.yml$
+^doc$
```
hadley, line 14: Why do you add these?

vignettes/from-base.Rmd
```diff
@@ -8,41 +8,85 @@ vignette: >
...
+  "str_subset(x, pattern)",                   "grep(pattern, x, value = TRUE)",
+  "str_to_lower(x)",                          "tolower(x)",
+  "str_to_title(x)",                          "tools::toTitleCase(x)",
+  "str_to_upper(x)",                          "toupper(x)",
+  "str_trim(x)",                              "trimws(x)",
+  "str_which(x, pattern)",                    "grep(pattern, x)",
+  "str_wrap(x)",                              "strwrap(x)"
+)
+
+# since downlit seems to omit tables, we manually link to pkgdown reference for HTML output
```
hadley, line 59: I don't think this is the problem — I think the problem is that `library(stringr)` isn't shown in the output, so downlit can't see it. If you put `library(stringr)` in a separate chunk that's shown in the output, downlit should just work. This should substantially simplify the dependencies.

.Rbuildignore
salim-b: To ignore the output generated by `devtools::build_vignettes()`.

vignettes/from-base.Rmd
salim-b: I'm not familiar with how downlit works internally (it seems quite magic!). But the table cell content is still not autolinked even when there's a visible `library(stringr)` in the output:

<details>

Screenshot 2022-12-06 at 22-44-49 From base R
</details>

I've tested a few combinations, none of them convinced downlit to autolink table cell contents. On the other hand, all code in text paragraphs is already properly linked **without** a visible `library(stringr)` chunk, so I think the issue lies elsewhere...

.Rbuildignore
hadley: Ah, that's a workflow that we no longer recommend.

vignettes/from-base.Rmd
hadley: Ok, in any case, I'd rather fix this upstream, rather than manually adding the links here.

.Rbuildignore
salim-b: So, you want me to remove the added ignore rules?

vignettes/from-base.Rmd
salim-b: Would be great if this could be fixed in downlit. https://github.com/r-lib/downlit/issues/67 is about a different issue though, right?

I've only added the links in the stringr column since I think they're more important when consulting this table. But I wouldn't mind if both columns were autolinked. 😄

Do you want me to remove the custom downlit-code?

hadley: Yes, please file a downlit issue, and remove the custom downlit code.

hadley:
Thanks!

Issue #488, opened 2022-12-28
Incorrect deprecation message for `str_view_all`


cgoo4:
```
> lifecycle::last_lifecycle_warnings()
[[1]]
<warning/lifecycle_warning_deprecated>
Warning:
`str_view()` was deprecated in stringr 1.5.0.
ℹ Please use `str_view_all()` instead.
---
Backtrace:
 1. stringr::str_view_all(...)
```

matthewjnield:
I noticed the deprecation warning for str_view_all() today as well, and have submitted pull request #490 to correct it.

Example:

#> library(stringr)
#> str_view_all(string = "abc", pattern = "b")
#> Warning: `str_view()` was deprecated in stringr 1.5.0.
#> ℹ Please use `str_view_all()` instead.
#>  [1] │ a\<b\>c

matthewjnield:
After merging pull request #490, the deprecation warning displayed when str_view_all() is called is now correct:

``` r
remotes::install_github(repo = "tidyverse/stringr", ref = "741300390b042889f5b60578ed7e0f9ce41963c5", quiet = TRUE)
library(stringr)
str_view_all(string = "abc", pattern = "b")
#> Warning: `str_view_all()` was deprecated in stringr 1.5.0.
#> ℹ Please use `str_view()` instead.
#> [1] │ a<b>c
```

PR #490, opened 2022-12-31
Fix str_view_all() deprecation warning

matthewjnield:
In 1.5.0, str_view() was improved and as a result str_view_all() [was made redundant and was deprecated.](https://www.tidyverse.org/blog/2022/12/stringr-1-5-0/#str_view)  However, the deprecation warning that is displayed when str_view_all() is called states that str_view() is deprecated and that str_view_all() should be used. This change corrects the warning to say that str_view_all() is deprecated and that str_view() should be used. Fixes issue #488.


hadley:
Thanks! Can you please re-run the tests and update the affect snapshot tests?

matthewjnield:
I updated the failing snapshot test for str_view_all(). All tests pass now.

hadley:
Thank you!

PR #491, opened 2022-12-31
Fix str_view_all() deprecation warning test

matthewjnield:
In 1.5.0, str_view() was improved and as a result str_view_all() [was made redundant and was deprecated.](https://www.tidyverse.org/blog/2022/12/stringr-1-5-0/#str_view)  However, the deprecation warning that is displayed when str_view_all() is called states that str_view() is deprecated and that str_view_all() should be used. This change corrects the warning to say that str_view_all() is deprecated and that str_view() should be used. Fixes issue #488. Added with #490 after forgetting to update snapshot tests.


matthewjnield:
I accidentally submitted this pull request from the wrong branch of my forked repository, which explains why it didn't become a part of #490. I will resubmit with the commit made to the appropriate branch.

PR #494, opened 2023-01-14
Fix bug in str_trunc()

UchidaMizuki:
I found a bug in `str_trunc()` where strings are not properly truncated when `width` is small, as shown below.
The bug seems to be caused by `str_sub("abc", 0, -1)` outputting `"abc"` instead of `""`.
I have fixed the bug by changing the behaviour of getting the tail of the strings.

``` r
library(stringr)

trunc <- function(direction, width) str_trunc(
  "This string is moderately long",
  direction,
  width = width
)

trunc("left", 3)
#> [1] "...This string is moderately long"
trunc("center", 3)
#> [1] "...This string is moderately long"
trunc("center", 4)
#> [1] "T...This string is moderately long"

str_sub("abc", 0, -1)
#> [1] "abc"
# => Not equal to ""
```

<sup>Created on 2023-01-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/trunc.R
+++ b/R/trunc.R
@@ -32,13 +32,19 @@ str_trunc <- function(string, width, side = c("right", "left", "center"),
     )
   }
 
+  str_sub_tail <- function(string, start) {
+    ifelse(start == 0,
+           "",
+           str_sub(string, start, -1))
+  }
+
   string[too_long] <- switch(side,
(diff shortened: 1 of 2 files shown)


UchidaMizuki:
Close this pull request temporarily as I'm starting to think that `stringi::stri_sub()` might be misbehaving.

PR #496, opened 2023-01-18
Improve `str_detect` documentation (#489)

### LINKED ISSUE #489: Documentation of `str_detect()` could be improved
Right now it's 
`#' @param negate If TRUE, return non-matching elements.`
This to me is suggesting that `str_detect()` is returning elements of string, which it is not.
Maybe it would be better to say `If TRUE, inverts the resulting boolean vector` or something like that.
I'm happy to do a PR if you like.

kylebutts:
Fixes #489

### DIFF
--- a/R/detect.R
+++ b/R/detect.R
@@ -21,7 +21,7 @@
 #'   [boundary()]. An empty pattern, "", is equivalent to
 #'   `boundary("character")`.
 #'
-#' @param negate If `TRUE`, return non-matching elements.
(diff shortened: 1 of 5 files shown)


hadley:
Thanks for doing this :smile: For future reference, you need `Fixes #489` to actually close an issue when a PR is merged.

hadley:
That text needs to go in the PR description not the title; you should use the title as a human readable description of what your PR does.

hadley:
Thanks!

Issue #512, opened 2023-06-04
str_trunc() incorrectly snips rhs-of-ellipsis for truncated strings


eauleaf:
**str_trunc** with param `side = 'left'` or `side = 'center'` returns incorrect result when ellipsis is the the same size or one larger than `width`. The rhs of the `ellipsis` gets appended without trunctation. 

``` r
datr <- c('', 'a', 'aa', 'aaa', 'aaaa', 'aaaaaaa')

stringr::str_trunc(datr, width=4, side = 'left',   ellipsis = "..") # correct
#> [1] ""     "a"    "aa"   "aaa"  "aaaa" "..aa"
stringr::str_trunc(datr, width=4, side = 'right',  ellipsis = "..") # correct
#> [1] ""     "a"    "aa"   "aaa"  "aaaa" "aa.."
stringr::str_trunc(datr, width=4, side = 'center', ellipsis = "..") # correct
#> [1] ""     "a"    "aa"   "aaa"  "aaaa" "a..a"
stringr::str_trunc(datr, width=3, side = 'left',   ellipsis = "..") # correct
#> [1] ""    "a"   "aa"  "aaa" "..a" "..a"
stringr::str_trunc(datr, width=3, side = 'right',  ellipsis = "..") # correct
#> [1] ""    "a"   "aa"  "aaa" "a.." "a.."
stringr::str_trunc(datr, width=3, side = 'center', ellipsis = "..") # should be c("","a","aa","aaa","a..","a..")
#> [1] ""           "a"          "aa"         "aaa"        "a..aaaa"      "a..aaaaaaa"
stringr::str_trunc(datr, width=2, side = 'left',   ellipsis = "..") # should be c("","a","aa","..","..","..")
#> [1] ""          "a"         "aa"        "..aaa"     "..aaaa"    "..aaaaaaa"
stringr::str_trunc(datr, width=2, side = 'right',  ellipsis = "..") # correct
#> [1] ""   "a"  "aa" ".." ".." ".."
stringr::str_trunc(datr, width=2, side = 'center', ellipsis = "..") # should be c("","a","aa","..","..","..")
#> [1] ""          "a"         "aa"        "..aaa"     "..aaaa"    "..aaaaaaa"
stringr::str_trunc(datr, width=1, side = 'left',   ellipsis = "..") # correct
#> Error in `stringr::str_trunc()`:
stringr::str_trunc(datr, width=1, side = 'right',  ellipsis = "..") # correct
[...]

PR #513, opened 2023-06-07
Corrected str_trunc() to pass correct 'left' and 'center' values to str_sub() when internal variable `width...` is 0 or 1 

eauleaf:
Correction should close Issue #512 and #494.

Note: in my tests, the #494 pull request also fixes the issue. (However, the #494 fix is ~10% slower, likely due to the internal function definition, in the standard case where no 'left' or 'center' replacements of zero length occur. #494 is ~5% faster in the edge case corrected here, likely due to the internal function returning an empty sting directly instead of snipping to an empty string.)

Below is the summary of the issue:

- when nchar(width) is 1 more than nchar(ellipsis), side 'center' fails to return correct result
- when nchar(width) is == to nchar(ellipsis), sides 'center' and 'left' fail to return correct result

Both cases happen because:
str_sub(string[too_long], -width..., -1))
becomes:
str_sub('characters', 0, -1))
which returns all 'characters' rather than the correct number

Example: 
c('', 'a', 'aa', 'aaa', 'aaaa', 'aaaaaaa') |> str_trunc(3, 'center', "..")

Incorrectly returns: 
[1] ""           "a"          "aa"         "aaa"        "a..aaaa"    "a..aaaaaaa"

Rather than: 
[1] ""    "a"   "aa"  "aaa" "a.." "a.."

### DIFF
--- a/R/trunc.R
+++ b/R/trunc.R
@@ -30,15 +30,21 @@ str_trunc <- function(string, width, side = c("right", "left", "center"),
     cli::cli_abort(
       "`width` ({width}) is shorter than `ellipsis` ({str_length(ellipsis)})."
     )
+  } else if (width... > 1) {
+    end_c <- end_l <- -1
+  } else if (width... == 1) {
+    end_c <- 0; end_l <- -1
+  } else {
+    end_c <- end_l <- 0
   }
 
   string[too_long] <- switch(side,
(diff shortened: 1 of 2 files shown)


hadley:
Duplicate of #515

PR #514, opened 2023-06-10
Add new `str_dedent` function

chrimaho:
One incredibly helpful function in Python is the [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) function. Under the hood, this function uses regex to strip any leading spaces, while maintaining any internal indentation within a chunk of code.

This addition here re-implements the same functionality using native R code.

I've ensured to include 4 different unit tests for the same.


arnaudgallou:
I'd also like to have a `dedent()` function in R. Two comments:

* Wouldn't it make sense to trim leading and trailing whitespaces in the output or add an argument to do so? That would be very useful when making character strings from multi-line text in R:
  ``` r
  str_dedent("
    This is a long sentence that starts on the first line,
      continues on the second,
        and ends on the third one.
  ")
  ```

* The function currently doesn't support the following situation (when python's `textwrap.dedent()` does):
  ``` r
  str_dedent("
      foo
    bar
  ")
  ```
  It's probably uncommon enough not to support it but that could be detailed in the function documentation.

lionel- (commented):
It'd be nice if this function was consistent with `glue::glue()`, at least in important cases, so users don't have to remember differences or look up documentation every time.

I think this means in particular trimming the trailing whitespace by default:

```r
glue::glue("\nfoo\n") |> unclass()
#> [1] "foo"
```

There are other subtle differences where it's probably not helpful to follow glue. For instance, the indentation is computed from all empty lines, including those that are trimmed, in a way I don't fully understand:

```r
glue::glue("\n \n\n ") |> unclass()
#> [1] "\n"
glue::glue("\n  \n\n ") |> unclass()
#> [1] " \n"
glue::glue("\n  \n\n  ") |> unclass()
#> [1] "\n"
glue::glue("\n  \n\n   ") |> unclass()
#> [1] "  \n"
```

I also don't understand the behaviour with 2 empty lines, I would have expected `""` here:

```r
glue::glue("\n") |> unclass()
#> [1] ""
glue::glue("\n\n") |> unclass()
#> [1] "\n"
glue::glue("\n\n\n") |> unclass()
#> [1] "\n"
glue::glue("\n\n\n\n") |> unclass()
#> [1] "\n\n"
glue::glue("\n\n\n\n\n") |> unclass()
#> [1] "\n\n\n"
```

For comparison (one more empty line than glue due to trailing one being preserved as of now):

```r
str_dedent("\n") |> unclass()
#> [1] ""
str_dedent("\n\n") |> unclass()
#> [1] "\n"
str_dedent("\n\n\n") |> unclass()
#> [1] "\n\n"
str_dedent("\n\n\n\n") |> unclass()
#> [1] "\n\n\n"
str_dedent("\n\n\n\n\n") |> unclass()
#> [1] "\n\n\n\n"
```

So while I think compatibility with glue in terms of deleting empty trailing lines would be nice for this function, full compat is probably not worth pursuing.

hadley:
@DavisVaughan you mean "Strips all leading and trailing __whitespace lines__  from the output" right? 

And you both really think we don't want a trailing new line? If you were going to `cat()` this you would want a trailing `\n`?

R/remove.R
```diff
@@ -19,3 +19,65 @@ str_remove <- function(string, pattern) {
...
+  lines <- str_split(string, fixed("\n"))
+  map_chr(lines, str_dedent_1)
+}
+
+str_dedent_1 <- function(lines, trim_empty_ends = TRUE) {
+  if (length(lines) <= 1) {
+    return(lines)
+  }
+
+  ws <- str_length(str_extract(lines, "^ *"))
```
jennybc, line 67:
```suggestion
  ws <- str_length(str_extract(lines, "^[ \t]*"))
```

Seems like we should also be thinking about tabs?

jennybc:
>  if you cat multiple times before returning, you need newlines in between them)

I guess that's when I would use `cli::cat_line()`.

We're talking a lot about glue, which stringr imports. Which makes me wonder ... why isn't `str_dedent()` just `glue::trim()`? 🤔

hadley:
That is a good question. If I replace the existing implementation with a direct call to `glue::trim()` then I get the following failures:

```
── Failure ([test-remove.R:12:3](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html#)): strips common ws ──────────────────────────────────────────────────────────────────────────────
str_dedent("  Hello\n    World") (`actual`) not equal to "Hello\n  World" (`expected`).

`lines(actual)`:   "Hello" "World"  
`lines(expected)`: "Hello" "  World"

── Failure ([test-remove.R:13:3](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html#)): strips common ws ──────────────────────────────────────────────────────────────────────────────
str_dedent("    Hello\n  World") (`actual`) not equal to "  Hello\nWorld" (`expected`).

`lines(actual)`:   "Hello"   "World"
`lines(expected)`: "  Hello" "World"

── Failure ([test-remove.R:25:3](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html#)): preserves final newline ───────────────────────────────────────────────────────────────────────
str_dedent("  Hello\n  World\n") (`actual`) not equal to "Hello\nWorld\n" (`expected`).

`lines(actual)`:   "Hello" "World"   
`lines(expected)`: "Hello" "World" ""

── Failure ([test-remove.R:35:3](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html#)): preserves final newline ───────────────────────────────────────────────────────────────────────
str_dedent("\n      Hello\n      World\n    ") (`actual`) not equal to "Hello\nWorld\n" (`expected`).

`lines(actual)`:   "  Hello" "  World"   
`lines(expected)`: "Hello"   "World"   ""
```

I can make most of them go away by adding a leading `\n` to the strings (which better reflects real use), but the remaining weirdness is this:

``` r
cat(
  glue::trim("
    Hello
      World
  ")
)
#>   Hello
#>     World
```

<sup>Created on 2025-09-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I find the extra indent here surprising. But maybe we could fix that?

PR #530, opened 2023-10-31
Improve split docs

hadley:


R/split.R
```diff
@@ -1,13 +1,20 @@
 #' Split up a string into pieces
 #'
 #' @description
-#' These functions differ primarily in their input and output types:
+#' This family of functions provide various ways of splitting a string up
```
jennybc, line 4:
```suggestion
#' This family of functions provides various ways of splitting a string up
```

R/split.R
```diff
@@ -1,13 +1,20 @@
...
 #'
-#' * `str_split()` takes a character vector and returns a list.
-#' * `str_split_1()` takes a single string and returns a character vector.
-#' * `str_split_fixed()` takes a character vector and returns a matrix.
-#' * `str_split_i()` takes a character vector and returns a character
-#'   vector.
+#' * `str_split_1()` takes a single string and splits it into pieces,
+#'    returning a single character vector.
+#' * `str_split_i()` splits each string in a character vector into pieces and
+#'    extracts the `i`th value, returning a character vector.
```
jennybc, line 10: This is not at all what I was asked to do / review but I'm still just noticing that these function names follow a common pattern. But then they do pretty different things. 🤔 I think this is the first time I've contemplated `str_split_i()`.

R/split.R
```diff
@@ -1,13 +1,20 @@
 #' Split up a string into pieces
 #'
 #' @description
-#' These functions differ primarily in their input and output types:
+#' This family of functions provide various ways of splitting a string up
+#' into pieces. There are two functions that return character vectors:
```
jennybc, line 5:
```suggestion
#' into pieces. These two functions return a character vector:
```

This is a nitpick, but I've started to think that using singular in these settings is much clearer.

R/split.R
```diff
@@ -1,13 +1,20 @@
...
-#' * `str_split_1()` takes a single string and returns a character vector.
-#' * `str_split_fixed()` takes a character vector and returns a matrix.
-#' * `str_split_i()` takes a character vector and returns a character
-#'   vector.
+#' * `str_split_1()` takes a single string and splits it into pieces,
+#'    returning a single character vector.
+#' * `str_split_i()` splits each string in a character vector into pieces and
+#'    extracts the `i`th value, returning a character vector.
+#'
+#' There are two functions that return more complex objects:
```
jennybc, line 12:
```suggestion
#' These two functions return a more complex object:
```

jennybc (approved):
Some weakly-held nitpicky opinions/suggestions.

R/split.R
hadley: Yeah, they both return a character vector, but for rather different reasons. Probably too later to do much about it now.

PR #544, opened 2024-04-10
`str_ilike()`

### LINKED ISSUE #543: SQL's LIKE is actually case sensitive
So need to change the default, and update the docs.

Probably worth a little exploration of impact on users (with github search for `str_like()`).

edward-burn:
- add str_ilike function
- deprecate ignore_case argument in str_like
- add tests of case sensitivity for both
- update NEWS.md
- update _pkgdown.yml

Fixes #543


R/detect.R
```diff
@@ -119,30 +119,69 @@ str_ends <- function(string, pattern, negate = FALSE) {
...
+                              what = "str_like(ignore_case)",
+                              details = "str_like() is always case sensitive. Use
+                              str_ilike() for case insensitive string matching.")
+  }
+
+  pattern <- regex(like_to_regex(pattern), ignore_case = FALSE)
+  stri_detect_regex(string, pattern, opts_regex = opts(pattern))
+}
+
+#' Detect a pattern in the same way as `SQL`'s `ILIKE` operator
```
hadley, line 154: I think you could combine the docs for these two functions. It'd also be useful to add a note about the incorrect behaviour of older versions of `str_like()`

R/detect.R
```diff
@@ -119,30 +119,69 @@ str_ends <- function(string, pattern, negate = FALSE) {
...
-str_like <- function(string, pattern, ignore_case = TRUE) {
+#' str_like(fruit, "%apple")
+str_like <- function(string, pattern, ignore_case = deprecated()) {
+  check_lengths(string, pattern)
+  check_character(pattern)
+  if (inherits(pattern, "stringr_pattern")) {
+    cli::cli_abort("{.arg pattern} must be a plain string, not a stringr modifier.")
+  }
+  if (lifecycle::is_present(ignore_case)) {
+    lifecycle::deprecate_warn(when = "1.5.2",
```
hadley, line 144: Could you please indent following the rules at https://style.tidyverse.org/syntax.html#long-lines?

tests/testthat/test-detect.R
```diff
@@ -57,9 +57,20 @@ test_that("functions use tidyverse recycling rules", {
 
 test_that("str_like works", {
   expect_true(str_like("abc", "ab%"))
+  expect_false(str_like("abc", "AB%"))
   expect_snapshot(str_like("abc", regex("x")), error = TRUE)
 })
 
+test_that("ignore_case is deprecated", {
+expect_snapshot(str_like("abc", "ab%", ignore_case = TRUE))
```
hadley, line 65:
```suggestion
  expect_snapshot(. <- str_like("abc", "ab%", ignore_case = TRUE))
```

tests/testthat/test-detect.R
```diff
@@ -57,9 +57,20 @@ test_that("functions use tidyverse recycling rules", {
 
 test_that("str_like works", {
```
hadley, line 58:
```suggestion
test_that("str_like is case sensitive", {
```

hadley (approved):
Thanks for working on this! It's a long term process to fix a bug like this and I appreciate you sticking with it 😄

R/detect.R
edward-burn: @hadley by this did you mean to use @inheritParams str_like? If so, I've added this here
https://github.com/tidyverse/stringr/pull/544/commits/1a366e3495422e1fec13495a2cd4bea58774baca But please let me know if I've misunderstood as I wasn't sure of what you meant

hadley: I meant using `@rdname`; see the changes I just pushed.

PR #550, opened 2024-07-15
Upkeep 2024 07

hadley:


krlmlr:
This still doesn't check Windows on R-devel. 🤷

hadley:
Does it need to?

krlmlr:
I've seen issues with stringi in the past, this motivated https://github.com/tidyverse/stringr/pull/548. We could also try to take this to the stringi repo, but as a user we would benefit from knowing if things go wrong.

PR #560, opened 2024-08-15
str_subset to keep names if present

### LINKED ISSUE #507: Doc clarification for str_subset about dropping names.
`str_subset` drops names. This appears to a design choice from `stringi` (https://github.com/gagolews/stringi/issues/59). Fair enough.

But I got bitten by this today, as I read the documentation stating: _"It's a wrapper around `x[str_detect(x, pattern)]`, and is equivalent to `grep(pattern, x, value = TRUE)`."_ But both those alternatives *do* keep names, and there is no other mention of dropping attributes. I suggest to add "but without preserving attributes like names", or something similar.

Consider:

```
fruit <- c(A = "apple", B = "banana", C = "pear", D = "pineapple")
str_subset(fruit, "b")
fruit[str_detect(fruit, 'b')]
grep('b', fruit, value = TRUE)
```

edward-burn:
closes #507

### DIFF
--- a/R/subset.R
+++ b/R/subset.R
@@ -31,12 +31,12 @@ str_subset <- function(string, pattern, negate = FALSE) {
   check_bool(negate)
 
   switch(type(pattern),
-    empty = no_empty(),
-    bound = no_boundary(),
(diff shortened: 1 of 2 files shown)


hadley:
Would be good to add a NEWS bullet in a separate PR 😄

thomasp85:
Thank you!

PR #563, opened 2024-08-15
Fix documentation that implies pattern argument can be empty

### LINKED ISSUE #508: Documentation for str_subset & str_detect imply pattern can be empty, but that gives an error
The documentation for str_detect and str_subset should be updated to say that pattern cannot be empty or boundary().

Currently both say "An empty pattern, "", is equivalent to boundary("character")", implying that they will accept a boundary() as the pattern, but this gives an error as below.

```
> fruit <- c("apple", "banana", "pear", "pineapple")
> str_subset(fruit, "")
Error in `str_subset()`:
! `pattern` can't be the empty string (`""`).

> fruit <- c("apple", "banana", "pear", "pineapple")
> str_subset(fruit, boundary("character"))
Error in `str_subset()`:
! `pattern` can't be a boundary.
```

wright13:
Fixes #508
Co-authored by: @RobLBaker

### DIFF
--- a/R/count.R
+++ b/R/count.R
@@ -4,6 +4,20 @@
 #' of `string.`
 #'
 #' @inheritParams str_detect
+#' @param pattern Pattern to look for.
+#'
+#'   The default interpretation is a regular expression, as described in
+#'   `vignette("regular-expressions")`. Use [regex()] for finer control of the
+#'   matching behaviour.
+#'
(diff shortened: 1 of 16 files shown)


hadley:
Thanks for this! I tweaked the wording a little and also added a similar caveat to `str_replace()` since that has the same limitation.

PR #564, opened 2024-08-15
add sep argument to str_dup

### LINKED ISSUE #487: Feature request: `sep` argument in `str_dup`.
I was going through this [SO question](https://stackoverflow.com/questions/74810135/is-there-an-r-function-that-will-repeat-a-number-limited-by-for-x-number-of), and thought that it might be useful to add an optional `sep` argument to `str_dup`.

For instance, in the case of the SO question the OP wants to repeat a string and add `;` between every repeated value.
```r
n = 3
input = c(67L, 24L, 72L, 3L)
expected_output = c("67;67;67", "24;24;24", "72;72;72", "3;3;3")
```
So far, what one can do is:
```r
library(stringr)
library(purrr)
map_chr(input, ~ str_flatten(rep(.x, each = n), ";"))
#[1] "67;67;67" "24;24;24" "72;72;72" "3;3;3"
```
Or in base R:
```r
sapply(input, \(x) paste(rep(x, n), collapse=";")) 
```
But what about a simple `str_dup` call?
```r
str_dup(input, n, sep = ";")
```

edward-burn:
closes #487

Adds a sep argument to str_dup. Becasue this is not supported by stri_dup, do separately and use lapply to go through strings.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # stringr (development version)
 
+* Add sep argument to `str_dup()` (@edward-burn, #564).
+
 * In `str_replace_all()`, a `replacement` function now receives all values in
   a single vector. This radically improves performance at the cost of breaking
   some existing uses (#462).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # stringr (development version)
 
-* Add sep argument to `str_dup()` (@edward-burn, #564).
+* Add sep argument to `str_dup()` so that it is possible to repeat a string and 
+  add a separator between every repeated value. (@edward-burn, #564).
 
 * In `str_replace_all()`, a `replacement` function now receives all values in
   a single vector. This radically improves performance at the cost of breaking
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # stringr (development version)
 
-* Add sep argument to `str_dup()` so that it is possible to repeat a string and 
-  add a separator between every repeated value. (@edward-burn, #564).
+* Add `sep` argument to `str_dup()` so that it is possible to repeat a string and 
+  add a separator between every repeated value (@edward-burn, #564).
 
 * In `str_replace_all()`, a `replacement` function now receives all values in
   a single vector. This radically improves performance at the cost of breaking
--- a/R/dup.R
+++ b/R/dup.R
@@ -11,6 +11,7 @@
 #' @examples
 #' fruit <- c("apple", "pear", "banana")
 #' str_dup(fruit, 2)
+#' str_dup(fruit, 2, sep = " ")
 #' str_dup(fruit, 1:3)
 #' str_c("ba", str_dup("na", 0:5))
 str_dup <- function(string, times, sep = NULL) {
@@ -21,7 +22,7 @@ str_dup <- function(string, times, sep = NULL) {
     stri_dup(input$string, input$times)
   } else {
     map_chr(seq_along(input$string), function(i) {
-      paste(rep(string[i], input$times[i]), collapse = sep)
+      paste(rep(string[[i]], input$times[[i]]), collapse = sep)
     })
   }
 }
--- a/R/dup.R
+++ b/R/dup.R
@@ -14,22 +14,14 @@
 #' str_dup(fruit, 1:3)
 #' str_c("ba", str_dup("na", 0:5))
 str_dup <- function(string, times, sep = NULL) {
-  vctrs::vec_size_common(string = string, times = times)
+  input <- vctrs::vec_recycle_common(string = string, times = times)
+  check_string(sep, allow_null = TRUE)
+
   if (is.null(sep)) {
-    stri_dup(string, times)
+    stri_dup(input$string, input$times)
   } else {
-    # stri_dup does not currently support sep
-    check_string(sep)
-    lapply(seq_along(string), function(i) {
-      if (vctrs::vec_size(times) == 1) {
-        paste(rep(string[i], times), collapse = sep)
-      } else {
-        paste(rep(string[i], times[i]), collapse = sep)
-      }
-    }) %>%
-      rlang::flatten_chr()
+    map_chr(seq_along(input$string), function(i) {
+      paste(rep(string[i], input$times[i]), collapse = sep)
+    })
   }
-
 }
-
-
--- a/R/dup.R
+++ b/R/dup.R
@@ -14,20 +14,22 @@
 #' str_dup(fruit, 1:3)
 #' str_c("ba", str_dup("na", 0:5))
 str_dup <- function(string, times, sep = NULL) {
-  vctrs::vec_size_common(string = string, times = times)
-  if(is.null(sep)){
+  size <- vctrs::vec_size_common(string = string, times = times)
+  if (is.null(sep)) {
     stri_dup(string, times)
   } else {
-  # stri_dup does not currently support sep
-  check_string(sep)
-  lapply(seq_along(string), function(i) {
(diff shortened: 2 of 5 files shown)


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # stringr (development version)
 
+* Add sep argument to `str_dup()` so that it is possible to repeat a string and
```
hadley, line 3:
```suggestion
* Add `sep` argument to `str_dup()` so that it is possible to repeat a string and 
```

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # stringr (development version)
 
+* Add sep argument to `str_dup()` so that it is possible to repeat a string and 
+  add a separator between every repeated value. (@edward-burn, #564).
```
hadley, line 4:
```suggestion
  add a separator between every repeated value (@edward-burn, #564).
```

tests/testthat/test-dup.R
```diff
@@ -13,3 +13,20 @@ test_that("0 duplicates equals empty string", {
 test_that("uses tidyverse recycling rules", {
   expect_error(str_dup(1:2, 1:3), class = "vctrs_error_incompatible_size")
 })
+
+test_that("uses sep argument", {
+  expect_equal(str_dup("a", 3, sep = ";"), "a;a;a")
+  expect_equal(str_dup("abc", 2, sep = "-"), "abc-abc")
+  expect_equal(str_dup("a", 3, sep = ";"), "a;a;a")
+  expect_equal(str_dup("abc", 2, sep = "-"), "abc-abc")
```
hadley, line 21: I think you could drop these tests because they're all basically executing all the same code.

tests/testthat/test-dup.R
```diff
@@ -13,3 +13,20 @@ test_that("0 duplicates equals empty string", {
...
+  expect_equal(str_dup(c("a", "b"), 2, sep = "x"), c("axa", "bxb"))
+  expect_equal(str_dup(c("aa", "bb", "ccc"), c(2, 3, 4), sep = ";"),
+               c("aa;aa", "bb;bb;bb", "ccc;ccc;ccc;ccc"))
+
+  expect_error(str_dup("a", 3, sep = 1))
+  expect_error(str_dup("a", 3, sep = c("-", ";")))
+  expect_error(str_dup(c("aa", "bb"), c(2, 3, 4), sep = ";"))
+  expect_error(str_dup(c("aa", "bb", "cc"), c(2, 3), sep = ";"))
```
hadley, line 29: Can you please use `expect_snapshot(error = TRUE)` here so I can see the error messages?

tests/testthat/test-dup.R
```diff
@@ -13,3 +13,20 @@ test_that("0 duplicates equals empty string", {
 test_that("uses tidyverse recycling rules", {
   expect_error(str_dup(1:2, 1:3), class = "vctrs_error_incompatible_size")
 })
+
+test_that("uses sep argument", {
+  expect_equal(str_dup("a", 3, sep = ";"), "a;a;a")
```
hadley, line 18: What happens if you do `str_dup(character(), 1, sep = ":")`?

R/dup.R
```diff
@@ -5,14 +5,29 @@
...
+str_dup <- function(string, times, sep = NULL) {
   vctrs::vec_size_common(string = string, times = times)
-  stri_dup(string, times)
+  if(is.null(sep)){
+    stri_dup(string, times)
+  } else {
+  # stri_dup does not currently support sep
+  check_string(sep)
+  lapply(seq_along(string), function(i) {
+    if (length(times) == 1) {
```
hadley, line 24: I don't love how we have to condition on the length of `times` to handle vectorisation (i.e. when times is a scalar and string is a vector). Could we instead take the output from the `vctrs::vec_size_common` and use that?

R/dup.R
```diff
@@ -5,14 +5,29 @@
...
-str_dup <- function(string, times) {
+str_dup <- function(string, times, sep = NULL) {
   vctrs::vec_size_common(string = string, times = times)
-  stri_dup(string, times)
+  if(is.null(sep)){
+    stri_dup(string, times)
+  } else {
+  # stri_dup does not currently support sep
+  check_string(sep)
+  lapply(seq_along(string), function(i) {
```
hadley, line 23: Could you use `map_chr()` here?

R/dup.R
```diff
@@ -5,14 +5,29 @@
...
 #' str_c("ba", str_dup("na", 0:5))
-str_dup <- function(string, times) {
+str_dup <- function(string, times, sep = NULL) {
   vctrs::vec_size_common(string = string, times = times)
-  stri_dup(string, times)
+  if(is.null(sep)){
+    stri_dup(string, times)
+  } else {
+  # stri_dup does not currently support sep
+  check_string(sep)
```
hadley, line 22: Great idea to check `sep` here.

hadley (commented):
Thanks for so much for your work. I've written a bunch of comments, but overall the direction is good, and I really appreciate you putting the time in. (And volunteering to be publicly eviscerated).

NEWS.md
edward-burn: fixed in latest commit

NEWS.md
edward-burn: fixed in latest commit

tests/testthat/test-dup.R
edward-burn: done in subsequent commit

tests/testthat/test-dup.R
edward-burn: changed these to snapshot tests in latest commit

tests/testthat/test-dup.R
edward-burn: added as a test. It returns character()

R/dup.R
edward-burn: I have tried this but the result of vctrs::vec_size_common is the length of the string input, not the length of times

hadley: Oh ooops needs to be `vec_recycle_common()`

R/dup.R
edward-burn: I might be being a bit dim, but I couldn't manage to get this to work. The problem I was hitting was that map_chr wants to return a length 1 vector.

edward-burn:
@hadley thanks for your review (and adding some nice comments in amongst the requested changes 😄). I have left a few comments above.

R/dup.R
```diff
@@ -5,14 +5,23 @@
...
 #' fruit <- c("apple", "pear", "banana")
 #' str_dup(fruit, 2)
 #' str_dup(fruit, 1:3)
 #' str_c("ba", str_dup("na", 0:5))
-str_dup <- function(string, times) {
-  vctrs::vec_size_common(string = string, times = times)
-  stri_dup(string, times)
+str_dup <- function(string, times, sep = NULL) {
+  input <- vctrs::vec_recycle_common(string = string, times = times)
+  check_string(sep, allow_null = TRUE)
```
hadley, line 18: I think it makes sense to move the other input check up here so that all the input checks are together.

tests/testthat/test-dup.R
```diff
@@ -13,3 +13,21 @@ test_that("0 duplicates equals empty string", {
 test_that("uses tidyverse recycling rules", {
   expect_error(str_dup(1:2, 1:3), class = "vctrs_error_incompatible_size")
 })
+
+test_that("uses sep argument", {
+  expect_equal(str_dup("abc", 1, sep = "-"), "abc")
```
hadley, line 18: I tried to make it a bit more clear was different in each test by keeping more stuff the same.

PR #566, opened 2024-08-15
Make type.character error for NA_character_

### LINKED ISSUE #546: Behavior of str_remove when called with NA_character_ pattern
When called with `NA` as a pattern, `str_remove` fails loudly about a type mismatch. When called with `NA_character_` as a pattern, it silently returns NA. Is this the expected behavior? If so it doesn't seem to be documented anywhere. Personally, I was expecting that `str_remove` would have returned the input string unmodified in this scenario (as there is nothing specified to remove), or alternatively failed with an error message.

``` r
library(stringr)

str_remove("test", NA)
#> Error in `str_replace()`:
#> ! `pattern` must be a string, not `NA`.

str_remove("test", NA_character_)
#> [1] NA

# What lead to discovery of this, though str_c is behaving as expected returning NA_character_

string_to_remove = NA # a blank cell in a data frame

str_remove_all("test", str_c("\\b",string_to_remove,"\\b"))
#> [1] NA
```

<sup>Created on 2024-04-11 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

nash-delcamp-slp:
Fixes #546

### DIFF
--- a/R/modifiers.R
+++ b/R/modifiers.R
@@ -202,15 +202,6 @@ type.stringr_fixed <- function(x, error_call = caller_env()) {
 }
 #' @export
 type.character <- function(x, error_call = caller_env()) {
-  if (length(x) == 1) {
-    if(is.na(x)) {
-      cli::cli_abort(
-        tr_("{.arg pattern} must be a string, not {.obj_type_friendly {x}}."),
-        call = error_call
-      )
-    }
-  }
-
   if (any(is.na(x))) {
     cli::cli_abort(
       tr_("{.arg pattern} must be a character vector that does not contain NAs."),
--- a/R/modifiers.R
+++ b/R/modifiers.R
@@ -220,7 +220,7 @@ type.default <- function(x, error_call = caller_env()) {
   }
 
   cli::cli_abort(
-    tr_("{.arg pattern} must be a string, not {.obj_type_friendly {x}}."),
+    tr_("{.arg pattern} must be a character vector, not {.obj_type_friendly {x}}."),
     call = error_call
   )
 }
--- a/R/modifiers.R
+++ b/R/modifiers.R
@@ -202,6 +202,22 @@ type.stringr_fixed <- function(x, error_call = caller_env()) {
 }
 #' @export
 type.character <- function(x, error_call = caller_env()) {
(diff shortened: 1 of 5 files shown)


R/modifiers.R
```diff
@@ -202,6 +202,22 @@ type.stringr_fixed <- function(x, error_call = caller_env()) {
...
 #' @export
 type.character <- function(x, error_call = caller_env()) {
+  if (length(x) == 1) {
+    if(is.na(x)) {
+      cli::cli_abort(
+        tr_("{.arg pattern} must be a string, not {.obj_type_friendly {x}}."),
+        call = error_call
+      )
+    }
+  }
```
hadley, line 212: Do you think this branch is really needed? I think just the error below would be sufficient.

R/modifiers.R
```diff
@@ -202,6 +202,22 @@ type.stringr_fixed <- function(x, error_call = caller_env()) {
...
+
+  if (any(is.na(x))) {
+    cli::cli_abort(
+      tr_("{.arg pattern} must be a string that does not contain NAs."),
```
hadley, line 216:
```suggestion
      tr_("{.arg pattern} must be a character vector that does not contain NAs."),
```

We try and use string to only mean "a character vector of length 1"

tests/testthat/test-modifiers.R
```diff
@@ -36,3 +36,15 @@ test_that("subsetting preserves class and options", {
...
+
+test_that("useful error message for NA_character_", {
+  expect_snapshot(error = TRUE, {
+    type(NA_character_)
+  })
+})
+
+test_that("useful error message for vector that includes NAs", {
+  expect_snapshot(error = TRUE, {
+    type(c("a", "b", NA_character_, "c"))
```
hadley, line 48: Great test!

hadley (commented):
Looking good! My main comment is that I think you can simplify your code a little. 

Could you please also add a news bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thanks for working on this! I hope you had fun at TDD 😁

PR #568, opened 2024-08-15
Adds a message when an empty string is provided

### LINKED ISSUE #497: Print something for `str_view(character())`
Should be adequate to fix just `print.stringr_view()` by adding a special case for `n = 0`.  I'm not sure what it should print though! Probably just something that it says "no strings", but more nicely 😄 

Will need a snapshot test.

LouisMPenrod:
Returns a message instead of returning nothing which was the previous behavior. Existing snapshot updated.

Fixes #497

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # stringr (development version)
 
+* `str_view()` now displays a message when called with a zero-length character
+  vector (@LouisMPenrod, #497).
 * In `str_replace_all()`, a `replacement` function now receives all values in
   a single vector. This radically improves performance at the cost of breaking
   some existing uses (#462).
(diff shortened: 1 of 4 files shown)


hadley:
@DavisVaughan do you remember what we've done when we've printed length-0 objects in other packages?

DavisVaughan:
You _could_ use the vctrs approach and always unconditionally print a header

```r
> vctrs::obj_print(1)
<double[1]>
[1] 1
> vctrs::obj_print(double())
<double[0]>
```

Like `<stringr_view[0]>` or something. So even in the empty case you get something.

---

But since this is mostly for interactive usage I don't mind something like `No strings provided` or something equally friendly. I don't really mind the current behavior much though, it is like `glue::glue()`.

Do we also want to consider the `str_view(character(), html = TRUE)` case?

hadley:
I consider `html = TRUE` basically superseded so no need to worry about it, I think.

hadley:
Thanks for working on this and I hope you had fun at TDD 😁

Issue #552, opened 2024-07-17
Convert `expect_error()` to `expect_snapshot()`


hadley:
Unless the test already uses `class`. 

Then verify that the error messages look ok, and make any fixes as needed

PR #570, opened 2024-08-15
Convert expect_error() to expect_snapshot()

KimLopezGuell:
Fixed issue #552

### DIFF
--- a/tests/testthat/test-conv.R
+++ b/tests/testthat/test-conv.R
@@ -6,7 +6,5 @@ test_that("encoding conversion works", {
 })
 
 test_that("check encoding argument", {
-  expect_snapshot(error = TRUE,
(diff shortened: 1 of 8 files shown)


tests/testthat/test-conv.R
```diff
@@ -6,5 +6,7 @@ test_that("encoding conversion works", {
 })
 
 test_that("check encoding argument", {
-  expect_error(str_conv("A", c("ISO-8859-1", "ISO-8859-2")), "single string")
+  expect_snapshot(error = TRUE,
```
hadley, line 9: When there's a single expression to test could you please fit on one line like this? 

```R
expect_snapshot(str_conv("A", c("ISO-8859-1", "ISO-8859-2")), error = TRUE)
```

hadley:
Thank you!

PR #577, opened 2024-12-10
Add functions: `str_to_pascal()`, `str_to_camel()`, `str_to_kebab()`, `str_to_snake()`

### LINKED ISSUE #573: Could a `str_to_snake` function be made to convert strings to snake-case?
It'd be useful to have a `str_to_snake` function in stringr.

I think the main use for this would be to rename column names to snake-case:

`rename_with(data, stringr::str_to_snake)`

It would enable not taking on another package dependency with `janitor::clean_names` in many user's workflows

librill:
Added 4 functions for common programming use. Closes #573 .

All functions call an existing `str_to_*()` so locale argument respects the behavior of those functions

Aliases `StrToPascal()` and `strToCamel()` exist (`str-to-kebab()` is not a valid name)

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -69,6 +69,7 @@ reference:
   - str_conv
   - str_like
   - str_replace_na
+  - str_to_camel
   - str_view
   - word
 


librill:
CC: @kbodwin

NAMESPACE
```diff
@@ -12,11 +12,13 @@ S3method(type,stringr_fixed)
 S3method(type,stringr_regex)
 export("%>%")
 export("str_sub<-")
+export(StrToPascal)
```
hadley, line 15: Now that I see them, I don't think it's worth having the camel case variants; it feels a bit too clever.

R/case.R
```diff
@@ -1,21 +1,41 @@
...
+#' Convert string to upper case, lower case, title case, sentence case, pascal
+#' case, camel case, kebab case, or snake case
 #'
 #' * `str_to_upper()` converts to upper case.
 #' * `str_to_lower()` converts to lower case.
 #' * `str_to_title()` converts to title case, where only the first letter of
 #'   each word is capitalized.
 #' * `str_to_sentence()` convert to sentence case, where only the first letter
 #'   of sentence is capitalized.
+#' * `str_to_pascal()` converts to pascal case, where only the first letter of
```
hadley, line 10: I'm not sure Pascal case is particularly well known (at least I didn't recognise it) so how about it make it an argument to `str_camel_case()`, e.g. `first_upper = TRUE`?

R/case.R
```diff
@@ -54,3 +74,50 @@ str_to_sentence <- function(string, locale = "en") {
...
+#' @export
+#' @rdname case
+strToCamel <- str_to_camel
+#' @export
+#' @rdname case
+str_to_snake <- function(string, separator = "_", locale = "en") {
+  stopifnot(is.character(string))
+  string <- string |>
+    str_replace_all("([a-z])([A-Z])", "\\1 \\2") |>
+    str_replace_all("([a-zA-Z])([0-9])", "\\1 \\2") |>
```
hadley, line 109: I think it would make sense to extract out this code for "break into words" into a separate function since it's reasonably complex and giving it a name would be useful.

R/case.R
```diff
@@ -54,3 +74,50 @@ str_to_sentence <- function(string, locale = "en") {
...
+  string <- str_to_pascal(string, locale = locale)
+  string <- str_replace(string, pattern = "^.", replace = str_to_lower(str_sub(string, 1, 1)))
+  return(string)
+}
+#' @export
+#' @rdname case
+strToCamel <- str_to_camel
+#' @export
+#' @rdname case
+str_to_snake <- function(string, separator = "_", locale = "en") {
```
hadley, line 105: You could eliminate the `separator` argument by creating an internal function that both snake and kebab case uses.

R/case.R
```diff
@@ -1,21 +1,41 @@
-#' Convert string to upper case, lower case, title case, or sentence case
+#' Convert string to upper case, lower case, title case, sentence case, pascal
```
hadley, line 1: Maybe it would be better to document these new functions in a new documentation topic? Yes, they're are changing case, but you're more likely to be apply them to function names, not regular sentences.

tests/testthat/test-case.R
```diff
@@ -10,6 +10,25 @@ test_that("to_title creates one capital letter per word", {
 })
 
 test_that("to_sentence capitalizes just the first letter", {
-  x <- "This is a sentence."
   expect_identical(str_to_sentence("a Test"), "A test")
 })
+
+test_that("to_pascal converts to pascal case", {
+  expect_identical(str_to_pascal("This is a sentence."), "ThisIsASentence")
```
hadley, line 17: I'd suggest testing cases the exercise each of the different regular expressions. I think that will be easier once you extract out a common "break into words" function.

PR #578, opened 2024-12-10
Converted stringr_view to S7.

### LINKED ISSUE #549: Try S7
Classes: 

* `stringr_pattern` subclasses in `modifiers.R`
* `stringr_view` in `view.R`

To tackle this task for TDD, you'll need some prior knowledge of S7. (But it's probably simple enough that you could pick up enough in one day to be effective, and if you try and don't succeed that would still be useful because it would indicate that the S7 documentation needs improvement).

VisruthSK:
Closes #549.

Created a new S7 class for `stringr_view`. Updated tests to respect S7 inheritance. Added S7 package as a dependancy.

Again mentioning @kbodwin as part of STAT 400!


hadley:
Thanks for working on this! I've finished it off, but looking at it now (especially with the warning that I've filed an issue about), it just doesn't look like it's worth making the switch for stringr yet.

Issue #575, opened 2024-11-01
str_sort drops vector names attribute


warnes:
With `stringr` 1.5.1, the `str_sort` function strips names from the vector.

Current behavior:
```r
x <- c(A="3", B="2", C="1")
> str(x)
 Named chr [1:3] "3" "2" "1"
 - attr(*, "names")= chr [1:3] "A" "B" "C"
> stringr::str_sort(x)
[1] "1" "2" "3"
> str(stringr::str_sort(x))
 chr [1:3] "1" 
```

Expected behavior:
```r
x <- c(A="3", B="2", C="1")
> str(x)
 Named chr [1:3] "3" "2" "1"
 - attr(*, "names")= chr [1:3] "A" "B" "C"
> stringr::str_sort(x)
[1] "1" "2" "3"
> str(stringr::str_sort(x))
 Named chr [1:3] "1" "2" "3"
 - attr(*, "names")= chr [1:3] "C" "B" "A"
```

-----------

```r
> sessionInfo()
R version 4.4.1 (2024-06-14)
Platform: x86_64-apple-darwin20
Running under: macOS 15.0.1

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.4-x86_64/Resources/lib/libRblas.0.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/4.4-x86_64/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.0

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: America/New_York
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

other attached packages:
[1] devtools_2.4.5 usethis_2.2.3

loaded via a namespace (and not attached):
 [1] vctrs_0.6.5       cli_3.6.2         rlang_1.1.4       stringi_1.8.4
 [5] purrr_1.0.2       pkgload_1.3.4     renv_1.0.11       promises_1.3.0
 [9] shiny_1.8.1.1     xtable_1.8-4      glue_1.7.0        htmltools_0.5.8.1
[13] httpuv_1.6.15     pkgbuild_1.4.4    ellipsis_0.3.2    fastmap_1.2.0
[17] lifecycle_1.0.4   memoise_2.0.1     stringr_1.5.1     compiler_4.4.1
[...]

PR #585, opened 2025-09-04
Preserve names where sensible

jonovik:
#575 points out that some stringr functions preserve names, but most don't. This pull request implements preservation of names where there should be no ambiguity.

#575 was closed with a recommendation to take the issue to stringi. I tried, and my pull request gagolews/stringi#521 was rejected as a matter of policy. Hence, I propose that we make this change in stringr. There are situations where preserving names is useful, and str_subset() and str_trunc() currently do that. In the stringi issue gagolews/stringi#59, @hadley argued that names is the only attribute that stringi should preserve.

This pull request makes focused modifications to the functions involved, and adds `tests/testthat/test-preserve-names.R`. All existing and new tests pass, and devtools::check() reports no errors or warnings, only a note which was already in current `main`.

Principles:

- Preserve names where the output has a 1-to-1 correspondence with the input.
- Preserve names for functions where base behavior suggests keeping names (e.g., grepl/grep(value=TRUE), sort/unique).
- Source of names: use names from the primary `string` argument only; ignore names on `pattern`/`replacement`/others and never merge.
- For 1-row-per-input matrices/lists (e.g., str_locate/str_match/str_split_fixed), set row/list names from input names.
- Don't preserve names where strings are combined, or the return values are indices
  (e.g., str_c, str_flatten, str_glue, str_which, str_order, str_equal).

These functions already preserve names:

- str_subset
- str_trunc

These functions preserve names if this pull request is accepted:

- str_count
- str_detect
- str_starts
- str_ends
- str_like
- str_escape
- str_replace
- str_remove
- str_conv
- str_trim
- str_pad
- str_sub
- str_to_lower
- str_to_upper
- str_to_title
- str_extract
- str_locate
- str_match
- str_extract_all
- str_locate_all
- str_match_all
- str_split
- str_split_fixed
- str_split_i
- str_sub_all
- str_length
- str_width
- str_dup
- word
- str_unique
- str_replace_na
- str_sort
- str_wrap
- str_replace_all

### DIFF
--- a/R/case.R
+++ b/R/case.R
@@ -28,29 +28,37 @@ NULL
 str_to_upper <- function(string, locale = "en") {
   check_string(locale)
 
-  stri_trans_toupper(string, locale = locale)
+  out <- stri_trans_toupper(string, locale = locale)
+  if (length(out) == length(string)) names(out) <- names(string)
+  out
 }
 #' @export
 #' @rdname case
 str_to_lower <- function(string, locale = "en") {
   check_string(locale)
 
-  stri_trans_tolower(string, locale = locale)
+  out <- stri_trans_tolower(string, locale = locale)
+  if (length(out) == length(string)) names(out) <- names(string)
+  out
 }
 #' @export
 #' @rdname case
 str_to_title <- function(string, locale = "en") {
   check_string(locale)
 
-  stri_trans_totitle(string, opts_brkiter = stri_opts_brkiter(locale = locale))
+  out <- stri_trans_totitle(string, opts_brkiter = stri_opts_brkiter(locale = locale))
+  if (length(out) == length(string)) names(out) <- names(string)
+  out
 }
 #' @export
 #' @rdname case
 str_to_sentence <- function(string, locale = "en") {
   check_string(locale)
 
-  stri_trans_totitle(
+  out <- stri_trans_totitle(
     string,
     opts_brkiter = stri_opts_brkiter(type = "sentence", locale = locale)
   )
+  if (length(out) == length(string)) names(out) <- names(string)
+  out
 }
--- a/R/extract.R
+++ b/R/extract.R
@@ -40,17 +40,21 @@
 #' str_extract_all("This is, suprisingly, a sentence.", boundary("word"))
 str_extract <- function(string, pattern, group = NULL) {
   if (!is.null(group)) {
-    return(str_match(string, pattern)[, group + 1])
+    out <- str_match(string, pattern)[, group + 1]
+    if (length(out) == length(string)) names(out) <- names(string)
+    return(out)
   }
 
   check_lengths(string, pattern)
-  switch(type(pattern),
+  out <- switch(type(pattern),
     empty = stri_extract_first_boundaries(string, opts_brkiter = opts(pattern)),
     bound = stri_extract_first_boundaries(string, opts_brkiter = opts(pattern)),
     fixed = stri_extract_first_fixed(string, pattern, opts_fixed = opts(pattern)),
     coll  = stri_extract_first_coll(string, pattern, opts_collator = opts(pattern)),
     regex = stri_extract_first_regex(string, pattern, opts_regex = opts(pattern))
   )
+  if (length(out) == length(string)) names(out) <- names(string)
+  out
 }
 
 #' @rdname str_extract
@@ -59,7 +63,7 @@ str_extract_all <- function(string, pattern, simplify = FALSE) {
   check_lengths(string, pattern)
   check_bool(simplify)
 
-  switch(type(pattern),
+  out <- switch(type(pattern),
     empty = stri_extract_all_boundaries(string,
       simplify = simplify, omit_no_match = TRUE, opts_brkiter = opts(pattern)),
     bound = stri_extract_all_boundaries(string,
@@ -71,4 +75,10 @@ str_extract_all <- function(string, pattern, simplify = FALSE) {
     regex = stri_extract_all_regex(string, pattern,
       simplify = simplify, omit_no_match = TRUE, opts_regex = opts(pattern))
   )
+  if (isTRUE(simplify)) {
+    if (is.matrix(out) && nrow(out) == length(string)) rownames(out) <- names(string)
+  } else {
+    if (is.list(out) && length(out) == length(string)) names(out) <- names(string)
+  }
+  out
 }
--- a/R/match.R
+++ b/R/match.R
@@ -54,10 +54,12 @@ str_match <- function(string, pattern) {
     cli::cli_abort(tr_("{.arg pattern} must be a regular expression."))
   }
 
-  stri_match_first_regex(string,
+  out <- stri_match_first_regex(string,
     pattern,
     opts_regex = opts(pattern)
   )
+  if (is.matrix(out) && nrow(out) == length(string)) rownames(out) <- names(string)
+  out
 }
 
 #' @rdname str_match
@@ -68,9 +70,11 @@ str_match_all <- function(string, pattern) {
     cli::cli_abort(tr_("{.arg pattern} must be a regular expression."))
   }
 
-  stri_match_all_regex(string,
+  out <- stri_match_all_regex(string,
     pattern,
     omit_no_match = TRUE,
     opts_regex = opts(pattern)
   )
+  if (is.list(out) && length(out) == length(string)) names(out) <- names(string)
+  out
 }
--- a/R/sub.R
+++ b/R/sub.R
@@ -64,11 +64,13 @@
 str_sub <- function(string, start = 1L, end = -1L) {
   vctrs::vec_size_common(string = string, start = start, end = end)
 
-  if (is.matrix(start)) {
+  out <- if (is.matrix(start)) {
     stri_sub(string, from = start)
   } else {
     stri_sub(string, from = start, to = end)
   }
+  if (length(out) == length(string)) names(out) <- names(string)
+  out
 }
 
 
@@ -88,9 +90,11 @@ str_sub <- function(string, start = 1L, end = -1L) {
 #' @export
 #' @rdname str_sub
 str_sub_all <- function(string, start = 1L, end = -1L) {
-  if (is.matrix(start)) {
+  out <- if (is.matrix(start)) {
     stri_sub_all(string, from = start)
   } else {
     stri_sub_all(string, from = start, to = end)
   }
+  if (is.list(out) && length(out) == length(string)) names(out) <- names(string)
+  out
 }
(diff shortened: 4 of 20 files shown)


tests/testthat/test-preserve-names.R
```diff
@@ -0,0 +1,147 @@
+x <- c(C = "3", B = "2", A = "1")
```
hadley, line 1: I think it would make more sense to put each test in the file corresponding to its function. And these tests should all be self-contained; i.e. it's better to redefine `x` inside of each of them so they are easy to run interactively.

R/case.R
```diff
@@ -28,29 +28,37 @@ NULL
 str_to_upper <- function(string, locale = "en") {
   check_string(locale)
 
-  stri_trans_toupper(string, locale = locale)
+  out <- stri_trans_toupper(string, locale = locale)
+  names(out) <- names(string)
+  out
```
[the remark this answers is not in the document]
jonovik: Thanks for the review.

A helper makes sense, but should we perhaps make `to` the "main" argument to `copy_names()` (as it is for `set_names()`)?
So the helper becomes

```
copy_names <- function(to, from) {
  set_names(to, names(from))
}
```

tests/testthat/test-preserve-names.R
jonovik: Will do.

R/extract.R
```diff
@@ -71,4 +75,10 @@ str_extract_all <- function(string, pattern, simplify = FALSE) {
     regex = stri_extract_all_regex(string, pattern,
       simplify = simplify, omit_no_match = TRUE, opts_regex = opts(pattern))
   )
+  if (isTRUE(simplify)) {
+    if (is.matrix(out) && nrow(out) == length(string)) rownames(out) <- names(string)
+  } else {
+    if (is.list(out) && length(out) == length(string)) names(out) <- names(string)
+  }
+  out
```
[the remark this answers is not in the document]
jonovik: It was indeed rather overwrought. Given what the stringi functions say about `simplify` (return matrix if true, else list) and the `check_bool()` before those lines, I have simplified it to:

```
  if (simplify) {
    rownames(out) <- names(string)
  } else {
    names(out) <- names(string)
  }
```

R/case.R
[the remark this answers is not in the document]
hadley: When copying, old normally comes first, not new.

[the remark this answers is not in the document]
jonovik: Okay, I'll update the PR.

R/match.R
```diff
@@ -68,9 +70,11 @@ str_match_all <- function(string, pattern) {
     cli::cli_abort(tr_("{.arg pattern} must be a regular expression."))
   }
 
-  stri_match_all_regex(string,
+  out <- stri_match_all_regex(string,
     pattern,
     omit_no_match = TRUE,
     opts_regex = opts(pattern)
   )
+  if (is.list(out) && length(out) == length(string)) names(out) <- names(string)
```
hadley, line 78: When is out not a list?

R/unique.R
```diff
@@ -26,5 +26,11 @@ str_unique <- function(string, locale = "en", ignore_case = FALSE, ...) {
     ignore_case = ignore_case,
     ...
   )
-  stri_unique(string, opts_collator = opts)
+  out <- stri_unique(string, opts_collator = opts)
```
hadley, line 29: I think you could maybe simplify this by using `stri_duplicated()` to subset `out`?

R/match.R
```diff
@@ -54,10 +54,12 @@ str_match <- function(string, pattern) {
     cli::cli_abort(tr_("{.arg pattern} must be a regular expression."))
   }
 
-  stri_match_first_regex(string,
+  out <- stri_match_first_regex(string,
     pattern,
     opts_regex = opts(pattern)
   )
+  if (is.matrix(out) && nrow(out) == length(string)) rownames(out) <- names(string)
```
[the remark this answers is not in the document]
jonovik: Always a matrix. Thanks, noted.

R/extract.R
```diff
@@ -59,16 +64,23 @@ str_extract_all <- function(string, pattern, simplify = FALSE) {
...
-      simplify = simplify, omit_no_match = TRUE, opts_fixed = opts(pattern)),
+      simplify = simplify, omit_no_match = TRUE, opts_fixed = opt),
     coll  = stri_extract_all_coll(string, pattern,
-      simplify = simplify, omit_no_match = TRUE, opts_collator = opts(pattern)),
+      simplify = simplify, omit_no_match = TRUE, opts_collator = opt),
     regex = stri_extract_all_regex(string, pattern,
-      simplify = simplify, omit_no_match = TRUE, opts_regex = opts(pattern))
+      simplify = simplify, omit_no_match = TRUE, opts_regex = opt)
   )
+  if (simplify) {
```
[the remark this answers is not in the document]
jonovik: Sorry, I missed your comment above. Let me have another go.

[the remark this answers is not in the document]
jonovik: I pushed 195159c with a proposal for keep_names() with tests. Could you please see if it is what you had in mind?

R/unique.R
jonovik: Now I'm not sure. Guarding against NA (logical) input to ensure character output means that this gets equally complicated whichever way we do it. The possible new revision below is equal in performance and memory use as the earlier one, but not much easier to read, nor -- I think -- more sophisticated in its string handling. We use base::match() only to find existing elements of `string`, so I think it should be fine.

Possible new revision:

```
str_unique <- function(string, locale = "en", ignore_case = FALSE, ...) {
  check_string(locale)
  check_bool(ignore_case)

  opts <- str_opts_collator(
    locale = locale,
    ignore_case = ignore_case,
    ...
  )

  # Ensure character output while preserving names of first occurrences
  string_chr <- as.character(string)
  keep <- !stringi::stri_duplicated(string_chr, opts_collator = opts)
  out <- string_chr[keep]
  names(out) <- names(string)[keep]
  out
}
```

My existing revision:

```
str_unique <- function(string, locale = "en", ignore_case = FALSE, ...) {
  check_string(locale)
  check_bool(ignore_case)

  opts <- str_opts_collator(
    locale = locale,
    ignore_case = ignore_case,
    ...
  )
  out <- stri_unique(string, opts_collator = opts)
  # Preserve names of first occurrence for each unique value
  if (!is.null(names(string))) {
    idx <- match(out, string)
    names(out) <- names(string)[idx]
  }
  out
}
```

jonovik: > Shouldn't that be "fixed" with `check_string()`?

The only other function currently doing `check_string(string` is `str_split_1()`. My impression was that in general, `stringr`'s principle is that `string` can be anything coercible to character. Thoughts?

R/utils.R
```diff
@@ -29,3 +29,40 @@ no_empty <- function(call = caller_env()) {
...
+#'
+#' @param out The result to potentially name; a vector, list, or matrix.
+#' @param string The primary input character vector whose names may be copied.
+#' @param pattern Optional pattern input used to decide if names should be
+#'   preserved.
+#' @return `out`, with `names(out)` or `rownames(out)` set from
+#'   `names(string)` when appropriate.
+#' @keywords internal
+#' @noRd
+keep_names <- function(out, string, pattern = NULL) {
```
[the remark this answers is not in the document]
jonovik: Many functions can then finish with `if (keep_names(string, pattern)) copy_names(string, out) else out`. I'll update the PR in a few minutes.

R/sub.R
```diff
@@ -64,11 +64,12 @@
 str_sub <- function(string, start = 1L, end = -1L) {
   vctrs::vec_size_common(string = string, start = start, end = end)
 
-  if (is.matrix(start)) {
+  out <- if (is.matrix(start)) {
     stri_sub(string, from = start)
   } else {
     stri_sub(string, from = start, to = end)
   }
+  if (length(out) == length(string)) copy_names(string, out) else out
```
hadley, line 72: Isn't `length(out) == length(string)` always going to be true?

R/sub.R
```diff
@@ -88,9 +89,10 @@ str_sub <- function(string, start = 1L, end = -1L) {
 #' @export
 #' @rdname str_sub
 str_sub_all <- function(string, start = 1L, end = -1L) {
-  if (is.matrix(start)) {
+  out <- if (is.matrix(start)) {
     stri_sub_all(string, from = start)
   } else {
     stri_sub_all(string, from = start, to = end)
   }
+  if (length(out) == length(string)) copy_names(string, out) else out
```
hadley, line 97: Same here

R/subset.R
```diff
@@ -30,13 +30,18 @@ str_subset <- function(string, pattern, negate = FALSE) {
...
-  switch(type(pattern),
-         empty = no_empty(),
-         bound = no_boundary(),
-         fixed = string[str_detect(string, pattern, negate = negate)],
-         coll  = string[str_detect(string, pattern, negate = negate)],
-         regex = string[str_detect(string, pattern, negate = negate)])
+  if (type(pattern) == "empty") no_empty()
+  if (type(pattern) == "bound") no_boundary()
 
+  idx <- str_detect(string, pattern, negate = negate)
```
hadley, line 36: I'd suggest keeping the original `switch` as this makes it easier to verify that every case is correctly handled.

R/subset.R
```diff
@@ -30,13 +30,18 @@ str_subset <- function(string, pattern, negate = FALSE) {
...
+  # str_detect() returns NA for NAs in string, but str_subset() should drop them
+  idx[is.na(idx)] <- FALSE
+  out <- string[idx]
+  # Work around the fact that as.character() drops names
+  nm <- names(out)
+  out <- as.character(out)
+  names(out) <- nm
```
hadley, line 43: Again, I don't think adding a new behaviour for a literal `NA` is worth complicating the code, and if you're going to convince me it needs to be in a separate PR not as an incidental change in this PR which is about preserving names.

R/unique.R
```diff
@@ -26,5 +26,11 @@ str_unique <- function(string, locale = "en", ignore_case = FALSE, ...) {
     ignore_case = ignore_case,
     ...
   )
-  stri_unique(string, opts_collator = opts)
+
+  # Ensure character output while preserving names of first occurrences
+  string_chr <- as.character(string)
```
hadley, line 31: Same here; leave literal `NA` inputs alone for now.

R/sub.R
jonovik: We could add `# Preserve names unless `string` is recycled` before the final line of `str_sub()` and likewise for `str_pad()`. But I see that the branch has now been merged. Would you still like me to (pull and) commit?

hadley: I just merged the updated main branch into your branch since I made a bunch of merge conflicts. You  can continue to edit it.

jonovik: @hadley, I added comments in 669a0e7 as described. Let me know if there is anything missing. Thanks!

PR #593, opened 2025-09-25
Refine string case conversion helpers

### LINKED ISSUE #592: Consider using unicode regex in `str_to_snake()` and friends
The new `str_to_snake()` and `str_to_kebab()` rely on non-unicode regex, which may not work in many languages:

``` r
stringr::str_to_snake("IlÉtaitUneFois")
#> [1] "ilétait_une_fois"

stringr::str_to_snake("ΜιαΦοράΚιΈνανΚαιρό")
#> [1] "μιαφοράκιένανκαιρό"

stringr::str_to_snake("ЖилиБыли")
#> [1] "жилибыли"
```

<sup>Created on 2025-09-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Could you consider using unicode regex instead?

arnaudgallou:
Fixes #592 

Let me know if you want me to add short comments explaining each regex pattern.

I'm not quite sure how you'd like to handle acronyms in `str_to_camel()`. I'm more inclined to treat acronyms like any other words (e.g. `userId`), but I know some people prefer preserving uppercase in two-letter acronyms (e.g. `userID`).

Some tests might be redundant.

### DIFF
--- a/R/case.R
+++ b/R/case.R
@@ -85,7 +85,7 @@ str_to_camel <- function(string, first_upper = FALSE) {
   string <- string |>
     to_words() |>
     str_to_title() |>
-    str_remove_all(pattern = "\\s+")
+    str_remove_all(pattern = fixed(" "))
 
   if (!first_upper) {
     str_sub(string, 1, 1) <- str_to_lower(str_sub(string, 1, 1))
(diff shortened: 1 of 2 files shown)


arnaudgallou:
I read a bit more on the differences between POSIX pre-built classes and unicode regex. The main difference is that POSIX classes are locale-dependent when unicode regex are not (see https://www.regular-expressions.info/posixbrackets.html).

I wasn't able to test it but POSIX classes may be able to handle the *ij* digraph properly if the locale is set to Dutch.

arnaudgallou:
Also, how would you like to handle superscript and subscript letters and numbers? Should we drop them, or normalize them using `stringi::stri_trans_nfkd()`? Normalizing the input string could be useful for handling certain units like `m²`, which also raises the question of how to treat digits.

Currently, digits are always separated from letters. Another option would be to keep digits attached when they directly follow a letter, but use the dash or underscore separator otherwise. The downside of this approach is that it makes the conversion from snake/kebab case to camel case non-reversible.

hadley:
Thanks! Lets keep this simple for now, and not worry about special characters (subscripts etc). 

I could also imagine options that ensure that the output is ASCII (since you might want that for programming variables) or valid R variable names, but I think extensions can wait until we see some use of this in the wild.

PR #608, opened 2026-06-07
tests: add edge-case tests for str_remove() and str_remove_all()

LeonidasZhak:
## Summary

Adds edge-case test coverage for `str_remove()` and `str_remove_all()`:
- **NA handling**: Tests that both functions correctly propagate `NA` values (single and mixed with non-NA)
- **Empty strings**: Tests behavior with empty string input
- **Zero-length vectors**: Tests that `character()` input returns `character()`
- **Factor input**: Tests factor-to-character coercion works correctly
- **No match**: Tests that non-matching patterns return the original string

## Motivation

`str_remove()` and `str_remove_all()` are thin wrappers around `str_replace()`/`str_replace_all()`, but the test file (9 lines) had zero coverage for these common data-cleaning edge cases. The DESCRIPTION claims all functions deal with NAs and zero-length vectors consistently — these tests verify that claim for the remove family.

## Tests run

```
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 15 ]
```

All 15 remove tests pass (5 new test blocks, 13 new assertions). Full suite has 2 pre-existing snapshot failures (lifecycle wording changes), unrelated to this work.

### DIFF
--- a/tests/testthat/test-remove.R
+++ b/tests/testthat/test-remove.R
@@ -7,3 +7,30 @@ test_that("str_remove() preserves names", {
   x <- c(C = "3", B = "2", A = "1")
   expect_equal(names(str_remove(x, "[0-9]")), names(x))
 })
+
+test_that("NA values pass through unchanged", {
+  expect_equal(str_remove(NA, "a"), NA_character_)
+  expect_equal(str_remove_all(NA, "a"), NA_character_)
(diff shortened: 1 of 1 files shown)


LeonidasZhak:
Withdrawing this small automated PR while I consolidate an oversized batch of contributions and reduce maintainer review burden. Sorry for the noise, and thank you for maintaining the project.

